PR #2, opened 2019-09-09
basic html structure for the website - Markus Schiffer

MarkusSchiffer:
This is my basic HTML framework for the website.


styles/desktop.css
```diff
@@ -0,0 +1,271 @@
+body {
```
MadEste: Looks good! 👍

styles/desktop.css
```diff
@@ -0,0 +1,271 @@
...
+  margin: 0 0 0.83rem 0;
+}
+h5 {
+  font-family: Georgia, sans-serif;
+  color: #213d69;
+  font-size: 1.5rem;
+  margin: 2rem 0 0.4rem 0;
+}
+h6 {
+  font-size: 0;
```
MadEste: Why the font size 0?

JavaScript/main.js
```diff
@@ -0,0 +1,12 @@
+let total = 0;
+
+function adopt(fee) {
+  total += fee;
+  alert(total);
+}
+
+function inspect(name, breed, fee) {
+  alert(name);
```
MadEste: It may be more preferable as a user to see one alert with a more nicely formatted message. 
See JS Template Literals:  

blog.html
```diff
@@ -0,0 +1,66 @@
...
+        <h6>Page footer</h6>
+        <img src="images/adoptadog-logo.svg" alt="Company dog logo">
+        <p>Aborpos nisto inci tet aut que volupti beruptiberro to ex esseque invenienihil imaximu scideru ptistiusam consequunt</p>
+        <nav id="bottom-nav">
+          <a href="index.html">Home</a> &nbsp;&nbsp;/&nbsp;&nbsp; <a href="dogs.html">Dogs</a> &nbsp;&nbsp;/&nbsp;&nbsp; <a href="#">Contact</a>
+        </nav>
+        <p>© 2018 Adoptadog, LLC | Seattle, WA</p>
+      </section>
+    </footer>
+    <script src="./JavaScript/main.js" charset="utf-8"></script>
```
MadEste, line 45: 👍

styles/desktop.css
MarkusSchiffer: Sorry for late response, but this header is for screenreader only.

styles/stylesheet.css
```diff
@@ -0,0 +1,380 @@
...
+  }
+  .divide {
+    display: flex;
+    justify-content: center;
+  }
+  .head-container {
+      padding-bottom: 4rem;
+      padding-top: 2rem;
+  }
+}
```
MadEste, line 280: You have a dead zone between 768 and 600 px where the aside becomes visible again and throws off the layout.

JavaScript/blog.js
```diff
@@ -0,0 +1,92 @@
...
+        ' oribus es quidignis ipietus explam sus am aut amet ant fugiatum,' +
+        ' utem non reptat.',
+        p2: 'Uptiusd andesci qui nem aut vendion ectur? Debis que explaut' +
+        ' laborenia que doluptur, con et labor abor sant poreperum dio quat' +
+        ' que doluptatur aut voluptassim quisciatquam ea ad qui con nes cus' +
+        ' esere dolut hicto teni solutenis alit ulparume nonseca estorer' +
+        ' spernam voluptassim quisciatquam ea ad qui con nes cus esere dolut' +
+        ' hicto teni solutenis alit ulparume nonseca estorer spernam.'
+    }
+];
```
MadEste, line 57: Good work with the array of objs

checkout.html
```diff
@@ -0,0 +1,172 @@
...
+      </div>
+    </header>
+    <nav class="top-nav">
+      <div class="standard-width">
+        <a href="index.html">Home</a> &nbsp;&nbsp;/&nbsp;&nbsp; <a href="dogs.html">Dogs</a> &nbsp;&nbsp;/&nbsp;&nbsp; <a href="blog.html">Blog</a>
+      </div>
+    </nav>
+    <section id="main-content" class="standard-width top-spacer">
+      <h2 class="occupy-solo">Checkout</h2>
+      <form method="GET" class="occupy-solo">
```
MadEste, line 32: Id suggest modifying the form fields font to better match the size and style of the page.

JavaScript/form.js
```diff
@@ -0,0 +1,39 @@
...
+    ];
+
+    let pass = true;
+
+    for (let i = 0; i < responses.length; i++) {
+        res = responses[i];
+        if (res == '') {
+            pass = false;
+        }
+        if (i === 1 && !responses[i].includes('@')) {
```
MadEste: This is a good start, but invalid addresses can still pass. ie : badEmail@badURL (no .com etc!)
There are many ways to validate, one being regex (Regular Expressions)
Here is one source I found on the subject: 
https://www.w3resource.com/javascript/form/email-validation.php
 But don't worry too much about it. regex is difficult even for experienced programmers and most people just google it!

MadEste: Small change, but I would instead do 
if !responses[1].includes('@')) { 
so there is only one check in the if and maybe move it outside of the loop so it doesn't  get checked every time. 
(very minor note)

JavaScript/main.js
```diff
@@ -0,0 +1,33 @@
...
+        const breed = "breed unknown";
+        const fee = "$" + $('a', this).data('fee');
+        alert('Name: ' + name + '\nBreed: ' + breed + '\nFee: ' + fee);
+    })
+    $('.adopt-button').click(function(e) {
+        e.stopPropagation();
+        if ($(this).data('availability') == "t") {
+            total += $(this).data('fee');
+            alert("$" + formatter.format(total));
+            $(this).data('availability', "f");
```
MadEste, line 26: Nice work using the data value.

PR #4, opened 2019-09-12
Four new html pages

jocjoh:
I have created four new html pages based on the mock-ups for this project, including a home page, a page that features bios and photos of all the adoptable dogs, a blog that offers tips and news-you-can-use, and a user friendly checkout form that allows us to get to know the customer. I think these changes are sustainable, user-friendly, and when partnered with responsive design, will make for a great user experience.


blog.html
```diff
@@ -0,0 +1,57 @@
...
+      <p><a href="checkout.html">Checkout</a></p>
+      <img src="images/adoptadog-logo.svg" alt="Adoptadog logo with dog in white" style="width:15%">
+      <h1>Seattle's premiere dog adoption agency</h1>
+      <nav>
+        <a href="index.html">Home</a> /
+        <a href="dogs.html">Dogs</a> /
+        Blog
+      </nav>
+    </header>
+    <h2>Adoptadog blog</h2>
```
MadEste: I'd consider moving into the section element to avoid the html validator warning. 
https://www.w3.org/wiki/HTML/Usage/Headings/Missing

javascript/java.js
```diff
@@ -0,0 +1,10 @@
+function dogInfo(name, breed, cost) {
+  alert('More on this dog:\n' + name + '\n' + breed + '\n' + cost + ' to adopt');
+}
+
+let total = 0;
+
+function totalCost(x) {
+  let newTotal = total += x;
+  alert('$' + newTotal);
+}
```
MadEste, line 13: Functionally this works, but in practice there is some wasted code here. For lines 8 and 9 I would change it to the following:
 ` total += x;`
  `alert('$' + total);`
The reason is the variable newTotal is sort of redundant, we already have a variable 'total' that we are using to keep track of the changes in price. newTotal only exists for an instant while the function runs (because it is block scoped or 'defined' inside the function) and is then removed from memory. 
Hopefully this will make more sense as we continue more Javascript exercises!

styles/stylesheet.css
```diff
@@ -0,0 +1,353 @@
...
+@media screen and (max-width: 991px) {
+  .flex-container-home {
+    width: 50%;
+    float: left;
+  }
+}
+
+@media screen and (max-width: 767px) {
+  .flex-container-home {
+    float: center;
```
MadEste: There is no 'center' option for float. Try L and R margins auto instead.

styles/stylesheet.css
```diff
@@ -0,0 +1,353 @@
...
+}
+
+@media screen and (max-width: 576px) {
+  .box-home {
+    width: 100%;
+    float: center;
+  }
+}
+
+.dog_photo {
```
MadEste, line 249: I think you are trying to make the dog images larger here. You need the container to be larger to give them more space. Edit the width of .flex-container-home instead

styles/stylesheet.css
```diff
@@ -0,0 +1,353 @@
+
+/*Style for header starts here*/
+header {
```
MadEste, line 7: Recomend setting body margins to 0, they have some default padding that is more noticeable at lower resolutions

styles/stylesheet.css
jocjoh: I'm confused because I fixed this after I ran my code through the validator. The "float: center" is no longer in my CSS so I'm not sure why it showed up here for you.

styles/stylesheet.css
jocjoh: Thank you, that seemed to take care of my wasted white space issue. Could you take a peek one more time to see if I resolved it? https://jocjoh.github.io/html200-adopt-a-dog/

styles/stylesheet.css
jocjoh: I tried setting the body margin to 0, but the right side is still showing a margin in the mobile viewport. Any ideas of what I'm doing wrong?

styles/stylesheet.css
```diff
@@ -0,0 +1,510 @@
...
+aside {
+  display: block;
+  float: right;
+  width: 20%;
+  margin-right: 15%;
+  font-size: 80%;
+}
+
+.aside_image {
+  width: 80%;
```
MadEste, line 149: I would consider changing this to 100% or else edit the margins for the p elements, some text is not lining up with the edges of the images for the aside.

PR #5, opened 2019-10-06
Pull for Assignment

KDMcDade:


blog.html
```diff
@@ -0,0 +1,75 @@
...
+        <a href="dogs.html">Dogs</a>  /
+        <a href="blog.html">Blog</a>
+      </p>
+    </nav>
+
+    <section>
+      <h1>Adoptadog Blog</h1>
+      <div>
+
+        <article>
```
MadEste: I would recommend using a div instead of article tag here. You have some warnings in your html validation as both sections and articles expect a header tag inside. Articles are meant to be self contained items that make sense on their own outside of the context of the page.

index.html
```diff
@@ -0,0 +1,165 @@
...
+            <p>Corrum volorit iandae nimaxim cum restia volor reicid ut et etur sunt arum  rendae pla endis re ea erum,  qui doluptae</p>
+            <button type="button" name="button" onclick="cost(123.45)"> Adopt</button>
+          </div>
+
+          <div class="doggo-grid">
+            <img src="images/poppy-card.jpg" alt="Poppy" onclick="poppy()">
+            <h3>Poppy</h3>
+            <p><strong>Cost to Adopt:</strong> $123.45</p>
+            <p>Corrum volorit iandae nimaxim cum restia volor reicid ut et etur sunt arum rendae pla endis re ea erum, qui doluptae</p>
+            <button type="button" name="button" onclick="cost(123.45)"> Adopt</button>
```
MadEste, line 75: Nice work with the onclick events.

dog.js
```diff
@@ -0,0 +1,57 @@
+
```
MadEste: This works but is not good code design since you have a unique function for each element.  Can you think of ways to modify this to have one function that accomplishes the same task? You could consider passing additional parameters in your function call in the html. Or you could create some key value pairs to store the message and price for each dog and use the name to locate and display. We will get more into these techniques in other classes!

dog.js
```diff
@@ -0,0 +1,57 @@
...
+function stella() {
+    alert('Stella, Smallo Doggo, $123.45')
+}
+
+let runningTotal = 0;
+
+function cost(price) {
+    let total = 0;
+    total += price;
+    runningTotal += price;
```
MadEste: Lines 53 & 54 are redundant since you are ultimately adding it to the variable runningTotal and the variable total is not being used.

styles/styles.css
```diff
@@ -0,0 +1,534 @@
...
+}
+
+/* Styling for Checkout page */
+.checkout {
+  color: #133e70;
+  padding: 2% 12%;
+  align-items: flex-start;
+  align-content: flex-start;
+  display: grid;
+  grid-template-columns: repeat(4, 1fr);
```
MadEste, line 356: This is affecting your css for the checkout button. grid columns seem unnecessary.

styles/styles.css
```diff
@@ -0,0 +1,534 @@
...
+/*Blog Stylings*/
+
+.blog-body {
+  padding: 1.5% 12% 0% 12%;
+  margin: auto;
+}
+
+.article {
+  display: grid;
+  grid-template-columns: repeat(2, 25%);
```
MadEste, line 333: articles are not taking up proper space because of 25% width

js/blog.js
```diff
@@ -0,0 +1,88 @@
...
+
+  //  Add to Blog Content Array
+  let blogArticles = [];
+
+  blogArticles.push(articlePostA);
+  blogArticles.push(articlePostB);
+  blogArticles.push(articlePostC);
+
+  //  Generate on page via for loop
+  for (let i = 0; i < blogArticles.length; i++) {
```
MadEste, line 86: This is good for appending them all with less code!

js/form.js
```diff
@@ -0,0 +1,20 @@
+//  Form Values  //
+const form = document.getElementById("checkout-form");
+const submitButton = document.getElementById("submit");
+
+let formInfo = [];
+
+function submitForm() {
+  let formInfo = []; //testing purposes//
```
MadEste: let variables are block scoped. your code will attempt to use the formInfo declared inside the function first.

styles/styles.css
```diff
@@ -0,0 +1,534 @@
...
+.main {
+  display: grid;
+  grid-template-columns: repeat(4, 1fr);
+  grid-gap: 10%;
+  padding: 3% 12%;
+  color: #1c356d;
+}
+
+.main-right {
+  grid-column: 3 / 4;
```
MadEste: The grid column here is not doing anything, you are only creating one column.

styles/styles.css
```diff
@@ -0,0 +1,534 @@
...
+.top-navigation> a:first-child {
+  padding: 0 1% 0 0;
+}
+
+
+/*Body*/
+
+.main {
+  display: grid;
+  grid-template-columns: repeat(4, 1fr);
```
MadEste: grid-template-columns: repeat( 4, minmax(25%, 1fr) );

Try something like the above to ensure they stay even sized.

js/dog.js
```diff
@@ -0,0 +1,22 @@
...
+  alert(`This is ${name}, they are the ${detail}. They cost ${price}.`);
+}
+
+//Shopping Cart//
+
+let runningTotal = 0;
+
+function cost(price) {
+    runningTotal += price;
+    $('.cart-amount').text("$" + runningTotal);
```
MadEste, line 13: 👍

PR #7, opened 2019-10-28
Add pages

Amyvanfo:
Add 4 pages of html to the project.


Amyvanfo:
Please add these pages to the project for adoptadog.

Amyvanfo:
found & fixed a problem with links. Updated. Please add this version to the project.

Amyvanfo:
Update the pages

Amyvanfo:
Updated for Assignment 6 Media Queries. Changes to files were mistakenly committed at html200-coding repository for this and the past couple of assignments, so some change commits are there.

styles.css
```diff
@@ -0,0 +1,297 @@
...
+@media screen and (max-width: 800px) and (min-width: 600px) {
+  body{background: white;}
+  .main, .main-dogs .footer, .header {
+    width: 100%;
+    margin-left: 10%;
+  }
+  .checkout {
+    top: 1rem;
+    right: 1rem;
+    position: fixed;
```
MadEste: Try not to use fixed. The button is not staying with the header now.

styles.css
```diff
@@ -0,0 +1,297 @@
...
+  }
+  .grid-container-2 {
+    display: grid;
+    grid-template-columns: repeat(2, 1fr);
+    grid-gap: 10px;
+    width: 80%;
+  }
+
+  aside {
+    width: 90%;
```
MadEste: I would play with this percentage, it seems to be cutting off elements at some sizes.

Amyvanfo:
Added background header, gradient and adjusted images so they are responsive.

styles.css
```diff
@@ -0,0 +1,372 @@
...
+    margin-top: none;
+    margin-left: -25%;
+  }
+  h1 {
+    font-size: 18px;
+    width: 25%;
+    word-wrap: normal;
+  }
+  .footer {
+    width: 800px;
```
MadEste: Do not use px for sizing as they are not responsive. Use % or em instead.

index.html
```diff
@@ -0,0 +1,165 @@
...
+          <img src="./images/violet-card.jpg" onclick="clickPhoto('Violet','Lab', '$123.45')" alt="brown lab puppy">
+          <h3>Violet</h3>
+          <p><b>cost to adopt:</b> $123.45</p>
+          <p>Corrum volorit iandae nimaxim cum restia volor reicid ut et etur sunt arum rendae pla endis re ea erum,qui doluptae</p>
+          <a class="button" onclick="dogFees(123.45)" >Adopt</a>
+          </div> -->
+      </div>
+    </section>
+    <div class="graybutton">
+      <a href="./dogs.html" style = "background-color: lightGray; color: darkblue; margin-top: 1rem; margin-left: 30rem;" target= "_blank"> See More Dogs</a><br>
```
MadEste: Try not to use inline styles, they make debugging and your styles difficult. 30 rem is far to high a number and is affecting other elements at smaller mobile resolutions.

Amyvanfo:
Thanks for catching these omissions. Oh boy.  I spent so much time trying
to figure out why I couldn't get the gray button to work in CSS.  I never
thought to look for CSS in the HTML doc.

Can I resubmit this for a better score?

Amy

styles.css
```diff
@@ -0,0 +1,370 @@
...
+    padding: 2px;
+    text-align: left;
+    background-color: white;
+    border: 1px;
+    border-color: lightGray;
+  }
+[class*="col-"] {
+  float: left;
+  margin-left: 2%;
+  margin-top: 1rem;
```
MadEste, line 153: I might recommend using either all % or all em units for margins, mostly just a coding style note.

styles.css
```diff
@@ -0,0 +1,370 @@
...
+    max-width: 90%;
+    min-width: 10%;
+    box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2);
+    padding: 2px;
+    text-align: left;
+    background-color: white;
+    border: 1px;
+    border-color: lightGray;
+  }
+[class*="col-"] {
```
MadEste, line 150: Your col classes could be making better use of their inner white space, there is alot of margin white space around the p and h2,h3 elements from the user stylesheets(default browser settings) try and explicitly set them so the look the same on all machines. 

By the same note, the body element has some margins being applied by the default browser style. I would set them to 0 explicitly.

styles.css
```diff
@@ -0,0 +1,370 @@
+
+* {
```
MadEste, line 2: Usually I set all elements to have 0 padding and margins at the top of the document. By targeting them all as you have here. You may find this changes your page greatly at first, but it helps you set each elements spacing explicitly so you can be sure it always looks the same for all users.

styles.css
```diff
@@ -0,0 +1,370 @@
...
+   }
+  .background {
+     width: 100%;
+     background-image: url("./images/header-image-narrow.jpg");
+   }
+   .footnav {
+     width: 90%;
+   }
+  header {
+    width: 30rem;
```
MadEste: I would set this to 100% so it takes up the whole screen instead of using rem.

styles.css
```diff
@@ -0,0 +1,370 @@
...
+  }
+  nav {
+    width: 30rem;
```
MadEste, line 281: same here

styles.css
```diff
@@ -0,0 +1,370 @@
...
+  margin-left: 16%;
+  padding-right: 1%;
+  float: left;
+  display: grid;
+  grid-template-columns: auto auto;
+}
+.blog-photo {
+  float: left;
+  clear: left;
+  padding: 1rem;
```
MadEste, line 175: All the cards seem to be resizing nicely! good work!

Amyvanfo:
Submitting Assignment 8.

dog-button.js
```diff
@@ -0,0 +1,92 @@
...
+  let text4 = $("#city").val();
+  let text5 = $("#zip").val();
+  let select = $("input:radio[name=first-time]:checked").val();
+  let state = $("#region").children("option:selected").val();
+  let location = $("#location").children("option:selected").val();
+  console.log('name:' + text + ' email: ' + text2 + ' Street address: ' + text3  + ' City: ' + text4  + ' Zip Code: ' +  text5 + 'first time: ' + select + ' state: ' + state + ' Pickup location: ' + location);
+  alert("Your form has been submitted");
+});
+
+const posts = [
```
MadEste: 👍

dog-button.js
```diff
@@ -0,0 +1,92 @@
...
+const postDiv = document.createElement('div');
+document.body.appendChild(postDiv);
+
+for (let i = 0; i < posts.length; i+= 1) {
+  let post = posts[i];
+
+  let blogTitle = document.createElement('h3');
+  let blogNode = document.createTextNode(post.title);
+  blogTitle.appendChild(blogNode);
+  document.getElementById("full-blog").appendChild(blogTitle);
```
MadEste: You are using the same file in both your form page and blog page. Because the form page does not have some elements you are getting errors when it cannot find element "full-blog". It is better practice to separate out your code into different files rather than load the same on both pages

dog-button.js
```diff
@@ -0,0 +1,92 @@
...
+/*hover effect on dog tiles*/
+$('.col-1').hover(function(e){
+  $(this).addClass('tile-border');
+})
+$('.col-1').mouseleave(function(e){
+  $(this).toggleClass('tile-border');
+})
+/*Handle form data on Checkout page*/
+$("form").submit(function(){
+  let text = $("#name").val()
```
MadEste: Very good! Only thing I would mention is your form is redirecting on submit. We don't want this to happen. 
use event.preventDefault() to keep this from happening. 

`$('form').submit(function (evt) {`
  ` evt.preventDefault(); //prevents the default action`

`}`

Amyvanfo:
This is submission for Assignment 9

blog-2.html
```diff
@@ -0,0 +1,54 @@
...
+        <img class= "footlogo" src="./images/adoptadog-logo.svg" alt="logo">
+        <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit</p>
+        <nav class="footnav">
+        <a href="index.html" style="background-color: rgb(17, 64, 105); color: white;"> Home /</a>
+        <a href="dogs.html" style="background-color: rgb(17, 64, 105); color: white;"> Dogs /</a>
+        <a href="blog.html" style="background-color: rgb(17, 64, 105); color: white;"> Blog </a>
+        </nav>
+          <p> 2018 Adoptadog, LLC | Seattle, WA </p>
+        </footer>
+<script src="https://code.jquery.com/jquery-3.4.1.js"></script>
```
MadEste, line 51: 👍

Amyvanfo:
This is submission for Assignment 10. Yay!

PR #1, opened 2019-08-04
Lesson 1 Pull Request - Shelby Goff

shelbygoff1:
This is the pull request for HTML200 Lesson One.

PR #10, opened 2019-11-30
Adding my html files

ahendryx:
I am adding my HTML files for Lesson #1.


styles/stylesheet.css
```diff
@@ -0,0 +1,374 @@
...
+
+nav {
+  background-color: #ccc;
+  margin-top: -16px;
+}
+
+ul {
+  list-style-type: none;
+  padding: 8px 0px 8px 0px;
+  margin-left: 60px;
```
MadEste: This margin is why it is not centering correctly. Target the UL for the menu separately so it doesn't apply to every ul on the page.

styles/stylesheet.css
```diff
@@ -0,0 +1,374 @@
...
+}
```
MadEste, line 75: I would also recommend using rem units instead of px for padding etc for a more scalable layout.

styles/stylesheet.css
ahendryx: Great. That really helps. Thank you!

scripts/script.js
```diff
@@ -0,0 +1,11 @@
+
+function showInfo(name,breed,fee) {
+  result = (name + ' is a ' + breed + ' and can be adopted for ' + fee + '.');
```
MadEste: Make sure you always declare a variable with 'let' 'const' etc... its generally bad practice to declare a variable in javascript without a keyword because it automatically becomes global and can lead to bugs.

scripts/script.js
```diff
@@ -0,0 +1,11 @@
...
+  alert(result);
+}
+
+
+function addTotal(fee) {
+  result = ('Your current total is $' + fee);
+  alert(result);
```
MadEste, line 13: Nice work using parameters to send data to the function. The key here is to keep update a persistent variable with the total. I would create a variable 'total' outside the function on line 1 : 
`let total = 0;`
then inside this function update it like so:
`total += fee;`
then report the total in alert.

styles/stylesheet.css
```diff
@@ -0,0 +1,561 @@
...
+button {
+  padding: 0px;
+  border: none;
+}
+
+.dog-profiles a {
+  font-family: 'Public Sans', sans-serif;
+  font-weight: 700;
+  font-size: .7em;
+  padding: 4px 60px 4px 50px;
```
MadEste: You are using px values for the button sizes, thus they do not scale appropriately different resolutions.

styles/stylesheet.css
```diff
@@ -0,0 +1,561 @@
...
+
+ul {
+  list-style-type: none;
+  padding: 8px 0px 8px 0px;
+  margin-left: 3rem;
+}
+
+li {
+  display: inline;
+  padding-right: 10px;
```
MadEste: Should not be using px units except for perhaps border lines

styles/stylesheet.css
```diff
@@ -0,0 +1,566 @@
...
+    width: 60%;
+    position: absolute;
+    top: 15%;
+    left: 4%;
+  }
+  .header h1 {
+    font-size: 0rem;
+  }
+  .header-checkout {
+    top: 50%;
```
MadEste, line 498: I'd be hesitant to place things using height%. On some resolutions this seems to be floating over content in the middle of the page.

styles/stylesheet.css
```diff
@@ -0,0 +1,566 @@
...
+  display: flex;
+  flex-flow: row wrap;
+  justify-content: start;
+  padding-top: 2rem;
+}
+
+.dog-profiles section {
+  border: 1px solid #ccc;
+  width: 25%;
+  margin: 0.8rem 3rem 3rem 0rem;
```
MadEste, line 181: Setting left and right margins to auto at smaller resolutions should center the dog tiles.

scripts/script.js
```diff
@@ -0,0 +1,23 @@
...
+let total = 0;
+function addTotal(fee) {
+  let result = ('Your total is $' + (total += fee));
+  alert(result);
+}
+
+
+/* Checkout Form */
+function submitInfo() {
+  let entries = document.getElementsByTagName('input', 'option');
```
MadEste: I don't think you can specify multiple tag types in one function call. You should be doing one for input, and another for select.
Be careful with radio buttons, they are of type input but report using .checked instead of .value
You could check with something like: 
`if (entries[i].getAttribute("type") == 'radio') {`
`console.log(`Radio Button checked? ${entries[i].checked} `);`
`}`

blog.html
```diff
@@ -0,0 +1,81 @@
...
+          <li><a href="blog.html">Blog</a></li>
+        </ul>
+      </nav>
+    </header>
+
+    <div class="container">
+      <article>
+        <h2 id="blog-h1">Title Goes Here</h2>
+          <div class="blog-article">
+            <section id="blog-1">
```
MadEste, line 39: Ideally you would not even have empty HTML skeleton elements to target. You should be able to look in the 
<div class="container">
      <article>
And create and insert new elements based on how much information is in your database or array/ object.

scripts/script-blog.js
```diff
@@ -0,0 +1,70 @@
+/* Add Header to Page */
+document.getElementById("blog-h1").innerHTML = "Adopt A Dog";
+
+
+
```
MadEste, line 5: So one thing you could try to do is create a js object or array of objects for each article, then you could use a loop to generate the same html content for each object and add it to the page. As far as adding things without a blank html element to target, the procedure is largely the same. Instead of 
`document.getElementById('blog-1');`
You can instead target your body element, and 
`.appendChild()` for each item. if you need unique id's for them you can use the iterator of the loop to create a unique string. If you need more explanation we can explain this better at office hours.

scripts/script.js
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    let b;
+    for (b = 0; b < entries2.length; b++) {
+    console.log(entries2[b].value);
+    }
+
+/*  if (entries1[a].getAttribute("type") == 'radio') {
+    console.log(Radio Button checked? ${entries1[a].checked} );
+  } */
```
MadEste: This looks ok except that it is not in the loop. I used something like 
`if(entries1[a].type == "radio"){`
   `     console.log("radio " + entries1[a].checked)`
      `}`

and it seemed to work, perhaps something else about your implementation was the issue?

PR #11, opened 2019-11-30
Adopt A Dog Assignment-1

sonalaggarwal555:
Lesson 2 Assignment 3 Adopt a Dog with CSS stylessheet along with html files


styles.css
```diff
@@ -0,0 +1,339 @@
...
+}
+.section-right-side-header{
+  color:darkblue;
+  font-family: serif;
+  font-size:x-large;
+}
+.section-right-side-para{
+  font-family: monospace;
+  padding: 0px 19px 0px 0px;
+  font-size:larger;
```
MadEste: While this is responsive I am not a huge fan of using the keywords for font-size as it can vary from browser to browser. % and em is more consistent.

styles.css
```diff
@@ -0,0 +1,339 @@
...
+  height:60px;
+}
+
+.card{
+display:grid;
+grid-template-rows: auto auto auto;
+border: 0.5px solid;
+}
+.card-image{
+ height: 100px;
```
MadEste: Images are not being sized properly at all resolutions. A better adjustment might be 
`height: auto;`
`width: 100%;`

styles.css
```diff
@@ -0,0 +1,339 @@
...
+}
+a{
+  color:white;
+  text-decoration: none;
+}
+header{
+  display: flex;
+  background-image: url('images/header-image.jpg');
+  height: 250px;
+  padding:40px;
```
MadEste: Adding padding to your header is throwing off the the dimensions for the rest of the page. Try disabling it and your navbar should now reach across the entire width.

Page-2.html
```diff
@@ -0,0 +1,310 @@
...
+          <h3>Here are all of our Dogs!</h3>
+        </section>
+
+        <article>
+        <!-- Row-1-->
+        <div class="row-1">
+          <div class="col">
+            <div class="card">
+                <div>
+                  <img src="images/murphy-card.jpg" class="card-image" alt="murphy" onclick="myFunction('Mruphy','abc','$123.45')"/>
```
MadEste: Good job linking onClicks

script.js
```diff
@@ -0,0 +1,19 @@
+
+var arr = new Array();
+function myFunction(name,breed,fee){
+  alert("Name:"+name+"\nBreed:"+breed+"\nAdoption Fee:"+fee);
```
MadEste, line 10: Nice work! May want to look at template literals for javascript string concatenation:

Page-2.html
```diff
@@ -0,0 +1,310 @@
...
+                </div>
+                <div class="card-desc">
+                  <h4>Murphy</h4>
+                  <p>Cost to Adopt: $<span id="price">123.45</span></p>
```
MadEste: You are using the id price multiple times. This is not good practice as html id's should be unique.

script.js
```diff
@@ -0,0 +1,19 @@
...
+}
+
+function getSum(total,num) {
+  return total + Math.round(num);
+}
+
+function adoptFunction(item){
+  var x = document.getElementById("price").childNodes[0].nodeValue;
```
MadEste: This always targets the first price element on your page and is a little redundant.

script.js
```diff
@@ -0,0 +1,19 @@
...
+  arr.push(x);
+
+  for (var i = 0; i <= arr.length; i++) {
+    total = +arr[i];
```
MadEste: Using a variable 'total' without ever defining it is bad practice. Should use let. using it in this way creates a global. 
Also I do not see you using total anywhere else so this loop seems a little unnecessary.

script.js
```diff
@@ -0,0 +1,19 @@
...
+  }
+  document.getElementById('demo').innerHTML = arr.reduce(getSum,0);
```
MadEste: This works! although keeping all the prices in an array and using the reduce function for calculation seems a bit of overkill.

index.html
```diff
@@ -0,0 +1,202 @@
...
+    <h4 class="nav-caption">
+      <a href="index.html"> Home </a>/
+      <a href="Page-2.html"> Dogs </a>/
+      <a href="Page-3.html"> Blog </a></h4>
+  </nav>
+<div class="container">
+  <!-- Main Container -->
+  <div class="main-container">
+    <section>
+      <div><img src="images/home-intro.jpg" alt="home"/></div>
```
MadEste: This image needs a percentage size, so it resizes correctly and does not interrupt your styles.

script.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const container = document.getElementById('container');
+const heading = document.createElement('h1');
+heading.textContent = 'Adoptadog Blog!';
+container.appendChild(heading);
+var images = ['blog-1.jpg','blog-2.jpg','blog-3.jpg'];
+var titles = ['Travelling With Your Dog','How To Walk Multiple Dogs','Teach Your Dog To Fetch!'];
+for(let i=0;i<images.length;i++){
```
MadEste: This is good, I would encourage you to think about how you could redo this to work without the individual sub-interface containers in the html

main.js
```diff
@@ -0,0 +1,65 @@
...
+  {
+    name:'first-time-adopter',
+    label:'First Time Adopter'
+  },
+  {
+    name:'location',
+    label:'Pickup location'
+  }
+];
+const form = document.createElement('form');
```
MadEste: Good Job using JS to generate the form!

main.js
```diff
@@ -0,0 +1,65 @@
...
+label.setAttribute('for', field.name);
+
+const input = document.createElement('input');
+input.setAttribute('id',field.name);
+input.setAttribute('type','text');
+input.setAttribute('value','');
+form.appendChild(label);
+form.appendChild(input);
+}
+let formId = document.getElementById('theForm').value;
```
MadEste: Form itself has no value. Input fields do.

main.js
```diff
@@ -0,0 +1,65 @@
...
+console.log('form:'+formId);
+container.appendChild(form);
+container.insertBefore(heading,form);
+
+const submitButton = document.createElement('button');
+submitButton.setAttribute('type', 'submit');
+submitButton.textContent = 'Submit';
+form.appendChild(submitButton);
+form.addEventListener('submit', function(e) {
+  alert('Thank you. The form information has been received');
```
MadEste: You are alerting on submit, but not reading any form data. You should be reading the .value of each form input field and reporting it. You can target the fields using the fields object you used to create the form.

styles.css
```diff
@@ -0,0 +1,406 @@
...
+  .row-1 {
+    display: grid;
+    grid-template-columns: auto !important;
+    grid-gap: 10px;
+    }
+  .col{
+    width: 50%;
+  }
+  .card-desc{
+    padding:2px !important;
```
MadEste, line 86: Should be getting away from px and instead use em or %

styles.css
```diff
@@ -0,0 +1,406 @@
...
+  }
+  .card-desc p{
+    padding: 0px !important;
+    font-size: 0.8 !important;
```
MadEste: must include a unit after the number. ex: 0.8em

MadEste: This error is causing your text to disappear at small breakpoints.

styles.css
```diff
@@ -0,0 +1,406 @@
...
+  p {
+    font-family: Calibri, Arial, Helvetica, sans-serif;
+  }
+  .container{
+    display:flex !important;
+    margin: 30px 50px;
+    padding:0px 150px;
+  }
+  .main-container{
+    flex:35% !important;
```
MadEste: this is being displayed as a block item, not flex so this property has no effect. 
https://www.w3schools.com/cssref/css3_pr_flex.asp

styles.css
```diff
@@ -0,0 +1,406 @@
...
+  nav{
+    width: 100% !important;
+    height: 25px !important;
+  }
+  body{
+    width: 100%;
+  }
+  .main-container{
+    margin: 3px !important;
+    grid-gap:5px !important;
```
MadEste: Again do not use px. The mockups for this assignment have gutters on either side of the page content. You should utilize some whitespace to either side of your content by setting the left and right margins to something like 10%

styles.css
```diff
@@ -0,0 +1,406 @@
...
+  }
+  h4{
+    padding: 2px !important;
+  }
+  nav{
+    width:100% !important;
+    height:25px !important;
+  }
+  .main-container{
+   margin: 3px !important;
```
MadEste, line 36: Plz no px. setting left and right margins to something like 10% would help create whitespace and center content.

styles.css
```diff
@@ -0,0 +1,406 @@
...
+    grid-gap:2px !important;
+    width: 100% !important;
+  }
+  .row-1 {
+    display: grid;
+    grid-template-columns: auto !important;
+    grid-gap: 10px;
+    }
+  .col{
+    width: 50%;
```
MadEste, line 83: This is shrinking the size of the dog cards. Remove this and let the main-container set the size of its contents.

PR #13, opened 2019-12-16
Lesson 1 - Semantic HTML - Elena 

ejdon:
Here are the pages with the semantic HTML added. 

Elena


stylesheets/desktop/stylesheet-adoptions.css
```diff
@@ -0,0 +1,212 @@
...
+#poppy {
+  grid-column: 2;
+  grid-row: 1;
+  border: 1px solid #ECECEC;
+}
+
+#jack {
+  grid-column: 3;
+  grid-row: 1;
+  border: 1px solid #ECECEC;
```
MadEste, line 63: One of the few places I think its ok to use px. but still be careful of how it looks at different device resolutions.

stylesheets/mobile/stylesheet-blog-mobile.css
```diff
@@ -0,0 +1,132 @@
+* {
```
MadEste: While its good to break up your css styles, have duplicates across different sheets can make it difficult to update your styles, since you might make a change on one sheet but forget to update the others. Better practice is to have one main stylesheet for common styles unique to all pages. (Think the header, footer, font family etc. )Then have seperate sheets like you have done, but only have styles for elements unique to that page, like the form or blog. then load multiple sheets. Remember that CSS loads top down and whatever is loaded last will overwrite whatever styles were loaded before.

stylesheets/desktop/stylesheet-blog.css
```diff
@@ -0,0 +1,132 @@
...
+  padding: 0% 1% 0 1%;
+}
+
+#blog-header {
+  color: #093C60;
+  font-size: 2em;
+  margin: 2% 0% 2% 15%;
+}
+
+#blog-1 {
```
MadEste: These all have the same styles, add a class to target them instead of an ID.

stylesheets/desktop/stylesheet-blog.css
```diff
@@ -0,0 +1,132 @@
...
+  display: flex;
+  flex-direction: row;
+}
+
+#blog-2 {
```
MadEste: At smaller resolutions the margins are gone since you applied them to the image but not the paragraph. it might be a better idea to add a margin bottom to the blog-#'s so they have some spacing between them.

index.html
```diff
@@ -0,0 +1,111 @@
...
+      </header>
+      <nav class="global-navigation">
+         <a href="index.html" class="nav-link">Home</a>
+         <p> / </p>
+         <a href="dogs.html" class="nav-link-2">Dogs</a>
+         <p> / </p>
+         <a href="blog.html" class="nav-link-2">Blog</a>
+      </nav>
+      <section class="index-content">
+         <div class="index-1">
```
MadEste: This div and index-2 are not resizing properly, they are being set by their content. Instead you should assign them some %width and then have their content resize to fill them. As is the dog image is not changing size and index-1 stays the same size while the text shrinks and has less size at smaller resolutions.

index.html
```diff
@@ -0,0 +1,111 @@
...
+            <img id="index-image" src="images/home-intro.jpg" alt="A picture of a man hugging a dog; the dog is smiling and looking into the camera">
+         </div>
+         <div class="index-2">
```
MadEste: In fact I might consider changing the html slightly to have a div container, then the img and the text as siblings. That way you could have the img take half the page by using width=50% and the text will automatically be side by side and wrap around underneath the img if the page becomes too small. ie:
`<div >`
`  `
` . <p> some text...</p>`
`</div>`

stylesheets/javascript/js.js
```diff
@@ -0,0 +1,14 @@
+
+/*Dog Info*/
+function dogInfo(name, breed, price) {
+  alert(`This is ${name}. They are a ${breed}. They cost ${price}.`);
+}
+
```
MadEste, line 6: Nice work using template literals for concatenation!

stylesheets/desktop/stylesheet-index.css
```diff
@@ -0,0 +1,158 @@
...
+aside {
+  width: 30%;
+  float: right;
+  margin-right: 3%;
+}
+
+.aside-blog-image {
+  margin-top: 3%;
+  width: 70%;
+  height: 12rem;
```
MadEste: Try not to mix units like this, (% and rem) you can see that some of your pics look distorted. Generally you want to set the width and leave the height blank so it will auto fill and keep the aspect ratio correct.

stylesheets/desktop/stylesheet-desktop-all.css
```diff
@@ -0,0 +1,96 @@
...
+  font-size: 1em;
+  font-weight: bold;
+  background: #ECECEC;
+  padding: 2% 0% 2% 0%;
+}
+
+.nav-link {
+  color: #093C60;
+  text-decoration: none;
+  padding: 0% 1% 0 15%;
```
MadEste: I think you want to use margins and not padding here, they are similar but can produce different results. Your nav-link is hiding behind nav-link-2 at some resolutions. using margins instead of padding should fix this.

stylesheets/javascript/blog-js.js
```diff
@@ -0,0 +1,105 @@
...
+    {
+        image: 'images/blog-3.jpg',
+        alt: 'The silhouette of a girl and a dog at sunset, playing fetch',
+        header: 'Teach Your Dog to Fetch!',
+        line1: 'Iduciendisite quo magnatem iuntum quid quaest ea am, tenderumet adis dolenem quidustrum fuga. Faceaquae estioria derum recuptatur, cum volore, undipsa doloreium hillupta aut es ut alitatuscit ommossum haritatur arum qui officae videbiti corporeium faccull oribus es quidignis ipietus explam sus am aut amet ant fugiatum, utem non reptat.',
+        line2: 'Uptiusd andesci qui nem aut vendion ectur? Debis que explaut laborenia que doluptur, con et labor abor sant poreperum dio quat que doluptatur aut voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam.'
+    }
+];
+
+for (let i = 0; i < blog.length; i += 1) {
```
MadEste, line 67: 👍 Nice work using a loop.

stylesheets/javascript/checkout-js.js
```diff
@@ -0,0 +1,13 @@
+function submitClick() {
```
MadEste, line 1: You seem to be using Jquery, you are required to do this for a later assignment but not #8. I have not looked ahead if you already submitted that or not. Nice work, my only addition would be to e.prevent default to keep the page from refreshing when you use the form. We do not want that behavior for this assignment.

PR #14, opened 2020-01-16
Lesson 1 assignment 

moreil25:
Pull request for Lesson 10 assignment, to include Read Me, testing, and User Personas.


styles/stylesheet.css
```diff
@@ -0,0 +1,403 @@
+/*colors
+navy #083c60
+green #81b741
+light gray #d3d3d3
+dark gray #3d3d3d
+*/
+
+* {
```
MadEste, line 8: I would consider setting the margins for all elements to 0 initially, this ensures any browser styles or overwritten and any margins are only the ones set by you.

styles/stylesheet.css
```diff
@@ -0,0 +1,403 @@
...
+}
+
+a, div {
+  font-family: 'Roboto', sans-serif;
+  text-decoration: none;
+}
+
+p {
+  font-family: 'Roboto', sans-serif;
+  font-size: 16px;
```
MadEste: Try not to use px for font or sizing anymore, they are not responsive, use em or % etc instead. Line border width is about the only place I would Use it.

index.html
```diff
@@ -0,0 +1,134 @@
...
+  <head>
+    <meta charset="utf-8">
+    <title>Home</title>
+    <link href="https://fonts.googleapis.com/css?family=Domine|Roboto&display=swap" rel="stylesheet">
+    <script src="https://kit.fontawesome.com/f399a12497.js" crossorigin="anonymous"></script>
+    <link rel="stylesheet" type="text/css" href="styles/stylesheet.css">
+  </head>
+
+  <body>
+  <a href="#main-content" class="hidden">Skip to main content!!</a>=
```
MadEste: the = is showing up in the page as it is not in the anchor tag.

scripts/scripts.js
```diff
@@ -0,0 +1,10 @@
+//Create click handlers that alerts the dog’s information when clicking on the dog’s photo.
+//You will need to pass in the params in the click handler of the dog’s name, breed and adoption fees.
+//Create click handlers on each dog’s button that will add to the total.
+//Alert this total.
+
+let x = myFunction();
+function myFunction () {
+  alert(//something);
```
MadEste: Not relevant for this assignment but you are getting a js error because you commented out  the entire line. 
This also commented out the parenthesis and every bracket type in code must be closed properly.

styles/stylesheet.css
```diff
@@ -0,0 +1,434 @@
...
+}
+
+.cart {
+  display: flex;
+  flex-flow: wrap row;
+  margin: 0;
+  /*margin: 1em;*/
+}
+
+button {
```
MadEste, line 367: I would modify this slightly to sit at the bottom and take up 100% of the width.

styles/stylesheet.css
MadEste: To add to this, by default chrome adds a margin of 8px to the body for me by default. Setting margins to 0 here would help make sure it takes up all the available space.

styles/stylesheet.css
```diff
@@ -0,0 +1,419 @@
...
+  grid-template-columns: auto auto auto;
+  margin: 1em;
+}
+
+.card {
+  width: 100%;
+  height: auto;
+}
+
+.dog-1, .dog-2, .dog-3, .dog-4, .dog-5, .dog-6, .dog-7, .dog-8, .dog-9, .dog-10, .dog-11, .dog-12 {
```
MadEste: It might make more sense to have one .dog class instead of unique ones. Usually if you need to target them individually you use an #id tag instead of .class

styles/stylesheet.css
```diff
@@ -0,0 +1,419 @@
...
+  padding: .25em;
+  border: none;
+  margin: 0;
+}
+
+.cost {
+  font-weight: 900;
+}
+
+nav ul {
```
MadEste, line 117: I might suggest adding some margin-bottom spacing to this element.

styles/stylesheet.css
```diff
@@ -0,0 +1,445 @@
...
+}
+
+.hidden:focus {
+  left: 0;
+}
+
+.header {
+  width: 100%;
+  height: 25em;
+  background: linear-gradient(to left, #d3d3d3, #3d3d3d, #083c60);
```
MadEste, line 26: 👍 nice work.

scripts/form.js
```diff
@@ -0,0 +1,34 @@
...
+$(document).ready(function () {
+  $('#buttonFinal').click(function myAdoption() {
+  const form = document.getElementById('buttonFinal');
+  form.addEventListener('submit', function(event) {
+    event.preventDefault();
+    myAdoption();
+    alert("Thank you. The form information has been received.");
+  });
+  });
+  $('#myForm').click(function() {
```
MadEste: It looks like you attached this to the form instead of the submit button, so it executes every time the form is clicked instead of on submission.

scripts/form.js
```diff
@@ -0,0 +1,34 @@
...
+submitButton.setAttribute('type', 'submit');
+submitButton.textContent = 'Submit';
+submitFinal.appendChild(submitButton);
+
+let total = 0;
+
+$(document).ready(function () {
+  $('#buttonFinal').click(function myAdoption() {
+  const form = document.getElementById('buttonFinal');
+  form.addEventListener('submit', function(event) {
```
MadEste: A submit event should be attached to the actual form element and not the button element. (by the way your #buttonFinal is actually targeting a div and not the button itself)
Buttons never trigger a submit event, so this code is never running.

PR #20, opened 2020-02-09
Lesson 3 Assignment 4

cdelapena:
Adopt-a-dog website with CSS and responsive design.


adoptadog-index.html
```diff
@@ -0,0 +1,182 @@
...
+      <nav class="nav-header">
+        <a class="nav-link" href="adoptadog-index.html">Home / </a>
+        <a class="nav-link" href="adoptadog-dogs.html">Dogs / </a>
+        <a class="nav-link" href="adoptadog-blog.html">Blog</a>
+      </nav>
+
+      <!--Home intro-->
+      <div class="grid-container">
+        <div class= "grid-item-3">
+          <img src="images/home-intro.jpg" alt="Introduction photo of dog and his owner" style="width:100%;">
```
MadEste, line 38: The Image and Text look good at larger resolutions but as the window shrinks you will notice a lot of wasted white space under the image in relation to text on the right. You can use breakpoints to change the layout and address this as we will cover later. You also might consider keeping the image and text as siblings in the same parent div and change the styles to achieve a similar look. This will allow the text to wrap and take up the white space at smaller resolutions.

styles/adoptadog-stylesheet.css
```diff
@@ -0,0 +1,154 @@
+header {
```
MadEste, line 6: You might consider adding 
`*{margin:0;}`
at the top of the style sheet. This trick removes any default browser styles and only applies margins explicitly set by your css. For me, Chrome always adds a margin to the body by default, ruining some of the header and footer styles.

cdelapena:
Closing lesson 3 pull request to create new one for lesson 4

PR #22, opened 2020-02-19
Lesson 4 assignment 2 pull request

cdelapena:


cdelapena:
Lesson 4 assignment 2 pull request

adoptadog-dogs.html
```diff
@@ -0,0 +1,231 @@
...
+
+      <h1 style="padding:0rem 4rem;">Here are all of our dogs!</h1>
+
+      <!--Create flexbox grid containing dog profiles-->
+      <div class="grid-container">
+        <div class= "grid-item">
+          <div>
+          <!--Use single quotations for strings-->
+          <button onclick="dogPhotoClick('Murphy', 'American Foxhound', 123.45)"><img src="images/murphy-card.jpg" alt="Picture of Murphy" style="width:100%;"></button>
+          <script src="./js/lesson4Assignment2.js"></script>
```
MadEste, line 41: You seem to have linked the JS several times after every button. You only need to link the file once. Usually it is best to link it at end of the body, that way the html loads first and then the script, just in case it references any html elements that were not yet loaded.

js/lesson4Assignment2.js
```diff
@@ -0,0 +1,18 @@
+function dogPhotoClick(dogName, dogBreed, adoptionFees) {
+
+  alert(dogName + ' is a(n) ' + dogBreed + '. Adoption cost is $' + adoptionFees + '.');
```
MadEste, line 3: Nice job, You can also use template literals for string concatenation: 

cdelapena:
Closing assignment 4 pull request to create new pull request for assignment 6.

PR #24, opened 2020-02-23
add html page only

tinawu1016:
Add html pages.

### DIFF
--- a/blog.js
+++ b/blog.js
@@ -1,0 +1,69 @@
+const footer = document.getElementById("blog-footer");
+
+const title = document.createElement("h2");
+title.textContent = "Adoptadog Blog";
+title.class = "container";
+title.className = "blog-title";
+document.body.insertBefore(title, footer);
+
+const container = document.createElement("div");
+container.className = "blog-container";
+document.body.insertBefore(container, footer);
+
+const blogs = [
+  {
+    imgSrc: "images/blog-1.jpg",
+    imgAlt: "A woman holds a black dog",
+    blogTitle: "Traveling With Your Dog",
+    p1: "Iduciendisite quo magnatem iuntum quid quaest ea am, tenderumet adis dolenem quidustrum fuga. Faceaquae estioria derum recuptatur, cum volore, undipsa doloreium hillupta aut es ut alitatuscit ommossum haritatur arum qui officae videbiti corporeium faccull oribus es quidignis ipietus explam sus am aut amet ant fugiatum, utem non reptat.",
+    p2: "Uptiusd andesci qui nem aut vendion ectur? Debis que explaut laborenia que doluptur, con et labor abor sant poreperum dio quat que doluptatur aut voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam.",
+
+  },
+  {
+    imgSrc: "images/blog-2.jpg",
+    imgAlt: "multiple dogs",
+    blogTitle: "How To Walk Multiple Dogs",
+    p1: "Iduciendisite quo magnatem iuntum quid quaest ea am, tenderumet adis dolenem quidustrum fuga. Faceaquae estioria derum recuptatur, cum volore, undipsa doloreium hillupta aut es ut alitatuscit ommossum haritatur arum qui officae videbiti corporeium faccull oribus es quidignis ipietus explam sus am aut amet ant fugiatum, utem non reptat.",
+    p2: "Uptiusd andesci qui nem aut vendion ectur? Debis que explaut laborenia que doluptur, con et labor abor sant poreperum dio quat que doluptatur aut voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam.",
+  },
+  {
+    imgSrc: "images/blog-3.jpg",
+    imgAlt: "A kid teach a dog to fetch",
+    blogTitle: "Teach Your Dog To Fetch!",
+    p1: "Iduciendisite quo magnatem iuntum quid quaest ea am, tenderumet adis dolenem quidustrum fuga. Faceaquae estioria derum recuptatur, cum volore, undipsa doloreium hillupta aut es ut alitatuscit ommossum haritatur arum qui officae videbiti corporeium faccull oribus es quidignis ipietus explam sus am aut amet ant fugiatum, utem non reptat.",
+    p2: "Uptiusd andesci qui nem aut vendion ectur? Debis que explaut laborenia que doluptur, con et labor abor sant poreperum dio quat que doluptatur aut voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam.",
+  }
+];
+
+
+for (let x = 0; x < blogs.length; x += 1) {
+    const object = blogs[x];
+
+    const blogImage = document.createElement("div");
+    blogImage.className = "blog-box1";
+    container.appendChild(blogImage);
+
+    const image = document.createElement("img");
+    image.setAttribute("src", object.imgSrc);
+    image.setAttribute("alt", object.imgAlt);
+    image.className = "blog-box1";
+    blogImage.appendChild(image);
+
+    const blogText = document.createElement("div");
+    blogText.className = "blog-box2";
+    container.appendChild(blogText);
+
+    const blogTitle = document.createElement("h2");
+    blogTitle.textContent = object.blogTitle;
+    blogText.appendChild(blogTitle);
+
+    const p1 = document.createElement("p");
+    p1.textContent = object.p1;
+    p1.className = "blog-intro";
+    blogText.appendChild(p1);
+
+    const p2 = document.createElement("p");
+    p2.textContent = object.p2;
+    p2.className = "blog";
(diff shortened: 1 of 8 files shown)


styles.css
```diff
@@ -0,0 +1,427 @@
+.header {
+  height: 330px;
```
MadEste: You should be using %, rem/em instead of px for most css from now on. px are not responsive.

styles.css
```diff
@@ -0,0 +1,427 @@
...
+.see-more a:active {
+  color: #fff;
+  background-color: #103861;
+}
+
+.blog-article {
+  display: flex;
+  flex-flow: row wrap;
+  margin: 0px 10px 40px 0px;
+  max-width: 90%;
```
MadEste: 👍 good.

styles.css
```diff
@@ -0,0 +1,427 @@
...
+}
+
+.blog-article img {
+  float: left;
+}
+
+.blog-text {
```
MadEste: Should be using margins instead of padding, margins are external while padding is internal. Also position absolute is not a great choice. since you are floating the image left, the text should normally appear right and take up remaining space.

styles.css
```diff
@@ -0,0 +1,427 @@
...
+
+.blog-article {
+  display: flex;
+  flex-flow: row wrap;
+  margin: 0px 10px 40px 0px;
+  max-width: 90%;
+}
+
+.blog-article img {
+  float: left;
```
MadEste: Make sure to add a width and height to the image to make sizing responsive.

styles.css
tinawu1016: Adjust all the px to rem or em.

styles.css
tinawu1016: I don't know what's happen to my code. If I don't use absolute, the content will go right, but under the picture. I tried so many ways, only the absolute can make the text stay next to the picture.

styles.css
tinawu1016: Added.

styles.css
```diff
@@ -0,0 +1,463 @@
...
+  color: #fff;
+}
+
+.header-checkout a:link {
+  text-decoration: none;
+  padding: 4px 8px 4px 8px;
+  background-color: #9bc344;
+  border-radius: 1px;
+  border: none;
+  width: 240px;
```
MadEste: Still want to be careful to use px for sizing or padding, fontsize etc. about the only place px is acceptable is defining border line width.

index.html
```diff
@@ -0,0 +1,140 @@
...
+        <div class="home-container">
+          <img src="images/home-intro.jpg" width="300" alt="Golden Retriever being hugged around the neck by its owner">
+          <h2>Adopt A Dog</h2>
+          <p>Iduciendisite quo magnatem iuntum quid quaest ea am, tenderumet adis dolenem quidustrum fuga. Faceaquae estioria derum recuptatur, cum volore, undipsa doloreium hillupta aut es ut alitatuscit ommossum haritatur arum qui officae videbiti corporeium faccull oribus es quidignis ipietus explam sus
+           am aut amet ant fugiatum, utem non reptat ad magnienecto volorent fugia nonsedi gendand itatquos as magnimpor aut molor autatia ad est officimus accae et omnia cus, od qui ate</p>
+        </div>
+
+        <div class="dog-profiles">
+          <section>
+            <img src="images/murphy-card.jpg" width="180" alt="Murphy">
```
MadEste: You should not have any inline styles, especially since this is interpreted by browsers as px and will overwrite any responsive css you have in your stylesheet.

styles.css
```diff
@@ -0,0 +1,463 @@
...
+  max-width: 90%;
+  padding-top: 2em;
+}
+
+.dog-profiles section {
+  border: .1em solid #ccc;
+  width: 10em;
+  margin: 1em 3em 2em 0em;
+}
+
```
MadEste, line 191: Add a .dog-profiles img { and set width to 100% so it resizes automatically to the container.

styles.css
```diff
@@ -0,0 +1,463 @@
...
+  background-color: #6a872c;
+}
+
+.dog-profiles a:active {
+  color: #fff;
+  background-color: #103861;
+}
+
+.dog-info {
+  padding: 1em 2em 3em 2em;
```
MadEste: I would remove the padding, it is affecting how content fits

styles.css
```diff
@@ -0,0 +1,463 @@
...
+  margin: 1em 3em 2em 0em;
+}
+
+.dog-profiles a:link {
+  font-family: sans-serif;
+  font-weight: 700;
+  font-size: .7em;
+  margin-top: 20px;
+  margin-bottom: 0px;
+  padding: 4px 70px 4px 62px;
```
MadEste: You should not be using px padding to set the size. 
You could instead try setting display to block and width to 100%

tinawu1016: Change the px problem.
I try to add display and width, but it turned out very wired, so I did not adjust this part.

styles.css
tinawu1016: Removed :)

styles.css
tinawu1016: changed.

index.html
tinawu1016: I moved it to css.

styles.css
tinawu1016: added

dog.js
```diff
@@ -0,0 +1,11 @@
+function showInfo(name, breed, cost) {
+  alert(name + ', ' + breed + ', '+ 'is cost '+ cost);
```
MadEste, line 2: 👍 This works well!

dog.html
```diff
@@ -0,0 +1,195 @@
...
+
+          <section>
+            <button onclick="showInfo('Luna', 'Mix', '$123.45')">
+              <img src="images/luna-card.jpg" class="dog-pic" alt="luna">
+            </button>
+             <div class="profile-info">
+              <h3>Luna</h3>
+              <p>Cost to Adopt: $123.45</p>
+              <p>Corrum volorit iandae nimaxim cum restia volor reicid ut et etur sunt arum rendae pla endis re ea erum, qui doluptae</p>
+              <a onclick="addTotal('123.45')">Adopt</a>
```
MadEste, line 162: this looks close, but since you are passing in 123.45 as a string it is not doing math correctly in your function.

dog.js
```diff
@@ -0,0 +1,11 @@
...
+}
+
+
+let total = 0;
+
+function addTotal(cost) {
+  total += cost;
```
MadEste: You need to make sure you are treating total as a Number, so Math works correctly.
This is called sanitizing your inputs:

`total += Number(cost);`

styles.css
```diff
@@ -0,0 +1,663 @@
...
+}
+
+
+.header-checkout img {
+  padding-right: 0;
+}
+
+
+.dog-pic {
+  width: 20em;
```
MadEste: Does not seem to scale correctly, instead consider using % so it is sized relative to the container. 

`Width: 100%;`

Consider something similar for fixing your button placement.

styles.css
```diff
@@ -0,0 +1,663 @@
...
+  }
+  .blog-preview {
+    display: flex;
+    flex-flow: column wrap;
+    border-left: 1px;
+    padding-left: 1rem;
+    padding-right: 1rem;
+    width: 25%;
+    float: right;
+    margin: 3rem 0 0 4rem;
```
MadEste, line 683: It looks like you are pushing the element off the page using margin. Instead use display:none; to actually hide it and make it take no space on the page.

styles.css
tinawu1016: Changed in PC document, will update later.

dog.js
tinawu1016: Got it. Changed in PC doc and will update later.

styles.css
tinawu1016: Got it. Adjusted.

dog.html
tinawu1016: I wonder know is there any way to do this part?

blog.js
```diff
@@ -0,0 +1,69 @@
...
+title.textContent = "Adoptadog Blog";
+title.class = "container";
+title.className = "blog-title";
+document.body.insertBefore(title, footer);
+
+const container = document.createElement("div");
+container.className = "blog-container";
+document.body.insertBefore(container, footer);
+
+const blogs = [
```
MadEste, line 13: Nice work using an array of objects!

blog.js
```diff
@@ -0,0 +1,69 @@
...
+    imgSrc: "images/blog-3.jpg",
+    imgAlt: "A kid teach a dog to fetch",
+    blogTitle: "Teach Your Dog To Fetch!",
+    p1: "Iduciendisite quo magnatem iuntum quid quaest ea am, tenderumet adis dolenem quidustrum fuga. Faceaquae estioria derum recuptatur, cum volore, undipsa doloreium hillupta aut es ut alitatuscit ommossum haritatur arum qui officae videbiti corporeium faccull oribus es quidignis ipietus explam sus am aut amet ant fugiatum, utem non reptat.",
+    p2: "Uptiusd andesci qui nem aut vendion ectur? Debis que explaut laborenia que doluptur, con et labor abor sant poreperum dio quat que doluptatur aut voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam.",
+  }
+];
+
+
+for (let x = 0; x < blogs.length; x += 1) {
```
MadEste, line 39: Good work using a loop to do the repetitive work.

checkout.js
```diff
@@ -0,0 +1,23 @@
...
+const street = document.getElementById("street");
+const city = document.getElementById("city");
+const state = document.getElementById("state");
+const first = document.getElementById("first-time-yes");
+const notFirst = document.getElementById("first-time-no");
+const pickup = document.getElementById("pickup");
+const submit = document.getElementById("submitForm");
+
+form.addEventListener("submit", function() {
+  event.preventDefault();
```
MadEste, line 6: 👍

PR #25, opened 2020-02-24
Creating pull request per Lesson 1 Assignment

HMalloy:
Probably did a lot more CSS work than was expected for this first lesson, I just can't help myself sometimes.


styles.css
```diff
@@ -0,0 +1,428 @@
+* {
+  box-sizing: border-box;
+}
+body {
+  margin:0;
+  font-family: sans-serif;
+  font-size: 18px;
```
MadEste: As you will learn later on on the course we are going to be getting away from using px since they are not responsive. Consider using % or em/rem instead.

HMalloy:
New for Lesson 2 assignment 3:
Updated HTML to use DIVs instead of Sections where appropriate
Went rem vs px crazy in the CSS
Added google fonts
Other refinements and details

script.js
```diff
@@ -0,0 +1,17 @@
+// dogInfo
+// addUp
+// #cart-total
+
+function dogInfo(name,breed,cost) {
+  alert('This is ' + name + ', they are a ' + breed + '. It is $' + cost + ' to bring them home with you!');
```
MadEste: Nice work! Consider using JS template literals for string concatenation. I think the syntax is nicer: 

script.js
```diff
@@ -0,0 +1,17 @@
...
+}
+
+let total = 0;
+
+function addUp(cost) {
+  total += cost;
+  alert('Total: $' + total);
+
+  // a little something extra = updates cart total in header, but only for the page you are on (for now)
+  document.getElementById("cart-total").innerHTML = total;
```
MadEste: 👍 Great work! this is exactly what you will be doing in future assignments.

script.js
HMalloy: This is great! Thanks!

script.js
```diff
@@ -0,0 +1,210 @@
...
+  {
+    title: 'Teach Your Dog To Fetch!',
+    image: 'images/blog-3.jpg',
+    snippet: 'Iciustiam, et ad quiaes cum que excea sinciendi qui offic temquas doluptios dolupta tiandebis nem ut.',
+    copy: `<p>Iduciendisite quo magnatem iuntum quid quaest ea am, tenderumet adis dolenem quidustrum fuga. Faceaquae estioria derum recuptatur, cum volore, undipsa doloreium hillupta aut es ut alitatuscit ommossum haritatur arum qui officae videbiti corporeium faccull oribus es quidignis ipietus explam sus am aut amet ant fugiatum, utem non reptat.</p>
+    <p>Uptiusd andesci qui nem aut vendion ectur? Debis que explaut laborenia que doluptur, con et labor abor sant poreperum dio quat que doluptatur aut voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam.</p>`
+  }
+];
+
+if (blog) {
```
MadEste, line 31: 👍 Nice job checking if there is content to load.

checkout.html
```diff
@@ -0,0 +1,126 @@
...
+        </nav>
+        <p>© 2018 Adoptadog, LLC | Seattle, WA</p>
+      </div>
+    </footer>
+    <script src="./script.js"></script>
+    <script>
+      const finish = document.getElementById('submitCheckout');
+
+      finish.addEventListener('submit', function(event){
+        event.preventDefault();
```
MadEste: 👍

PR #28, opened 2020-03-05
Godwin: adopt-a-dog

Headieh:
HTML code and some css for the mockups for assignment 1 (lesson1 assignment - 
852c13a)


Headieh:
HTML code and css for the mockups for lesson 2 assignment 3 (lesson2 assignment3 - 
0a6b9e2)

scripts/adopt.js
```diff
@@ -0,0 +1,60 @@
+function disableButton(id) {
+	document.getElementById(id).disabled = true;
```
MadEste, line 2: Nice concept to keep the user from pressing adopt multiple times on the same dog!

scripts/adopt.js
```diff
@@ -0,0 +1,60 @@
...
+}
+
+let eachFee = [];
+
+function adopt(x) {
+	if (x === 'murphy') {
```
MadEste, line 28: Why not pass the fee instead of name, and avoid the if statements?

scripts/about.js
```diff
@@ -0,0 +1,76 @@
+function about(x) {
```
MadEste: This works well, but remember you can also pass more than one parameter and avoid the hardcoded if statements.

styles/stylesheet.css
```diff
@@ -0,0 +1,303 @@
...
+}
+
+a {
+	color: rgb(81, 156, 183);
+	font-weight: bold;
+	display: block;
+	text-decoration: none;
+}
+
+a:hover,
```
MadEste, line 159: I believe this is causing some issues with your menu items "Jitter" on rollover.

scripts/checkout.js
```diff
@@ -0,0 +1,37 @@
...
+	info.push($("#city").val());
+	info.push($("#state").val());
+	info.push($("#zip").val());
+	info.push($("input:radio[name='New']:checked").val());
+	info.push($("#location").val());
+	console.log(info);
+	alert("Thank you. The form information has been received");
+}
+
+$('form').on('submit', submitEvent);
```
MadEste, line 18: Would prefer you accomplish this without JQuery for assignment 1, but everything looks good!

PR #29, opened 2020-03-22
Lesson 3 Assignment 4

paulfleurdelys:
Completed Adopt-a-Dog website HTML for Lesson 1 Assignment.  Includes 4 pages, homepage/index, dogs list, blog, checkout.


paulfleurdelys:
Lesson 02, Assignment 3, adding completed CSS for Adopt-a-Dog project.

styles/stylesheet.css
```diff
@@ -0,0 +1,391 @@
...
+}
+
+.header-nav{
+    background-color: #dad8d2;
+    color: #113c60;
+    font-weight: bold;
+    display: flex;
+    flex-direction: row;
+    justify-content: flex-start;
+    padding: 0px 210px;
```
MadEste: You should no longer be using px for padding, font size, etc. Use em, rem or % instead.

styles/stylesheet.css
```diff
@@ -0,0 +1,391 @@
...
+    font-size: 1.5rem;
+}
+
+input {
+    height: 1.5rem;
+}
+
+
+select, option {
+    height: 28px;
```
MadEste: try to use a different unit than px for height.

scripts/script.js
```diff
@@ -0,0 +1,10 @@
+function dogCardTotalOnClick(name, breed, dogPrice) {
+    alert("This dog's name is " + name + " and its breed is " + breed + " and the adoption fees are: " + dogPrice);
+}
+
+let cartTotal = 0;
```
MadEste: 👍

index.html
```diff
@@ -0,0 +1,148 @@
...
+                    <h1 class="header-logo-heading">Seattle’s premiere dog adoption agency</h1>
+                </div>
+                <div class="header-checkout-wrapper">
+                    <img class="header-checkout-icon" src="images/cart-icon.svg" alt="checkout cart icon">
+                    <p class="header-checkout-cart">$1234.56</p>
+                    <a href="checkout.html" class="header-checkout-link">Checkout</a>
+                </div>
+            </div>
+            <div class="header-background-image-container">
+                <img class="header-background-image" src="images/header-image.jpg" alt="dog eyes close up">
```
MadEste: I would suggest having the image as part of the header instead of using z index to cover up parts that do not fill the screen. It also needs to fit more of the width, as in the mockups. You will have more direction on how to do this in the next assignment using css instead of html elements.

styles/stylesheet.css
```diff
@@ -0,0 +1,555 @@
...
+    font-size: 1.5em;
+    margin: 1em 0px 0px 0px;
+}
+
+.blog-post-text > p {
+    line-height: 110%;
+}
+
+.blog-post-image {
+    object-fit: contain;
```
MadEste, line 358: I would consider adding width:100% so that the images scale as the page shrinks, and do not affect the header.

blog.html
```diff
@@ -0,0 +1,53 @@
...
+            
+            <nav class="header-nav">
+                <a href="index.html">Home</a>
+                <span>/</span>
+                <a href="dogs.html">Dogs</a>
+                <span>/</span>
+                <a href="blog.html">Blog</a>
+            </nav>
+        </header>
+    <main id="blog-main">
```
MadEste, line 35: 👍

scripts/script.js
```diff
@@ -0,0 +1,114 @@
...
+    {image_url:"images/blog-3.jpg", image_alt: "woman with dog playing fetch at sunset",title: "Teach Your Dog To Fetch!", text: ["Iduciendisite quo magnatem iuntum quid quaest ea am, tenderumet adis dolenem quidustrum fuga. Faceaquae estioria derum recuptatur, cum volore, undipsa doloreium hillupta aut es ut alitatuscit ommossum haritatur arum qui officae videbiti corporeium faccull oribus es quidignis ipietus explam sus am aut amet ant fugiatum, utem non reptat.", "Uptiusd andesci qui nem aut vendion ectur? Debis que explaut laborenia que doluptur, con et labor abor sant poreperum dio quat que doluptatur aut voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam."]}
+]
+
```
MadEste: Good use of object arrays to store data. You may want to declare with const rather than let for these purposes.

scripts/script.js
```diff
@@ -0,0 +1,114 @@
...
+function dogCardTotalOnClick(name, breed, dogPrice) {
+    alert("This dog's name is " + name + " and its breed is " + breed + " and the adoption fees are: " + dogPrice);
+}
+
+function dogCardPriceOnClick() {
+    alert ("This dog's price is $" + this.cost);
+}
+
+function createDogCards() {
+    dogCards.forEach(e => {
```
MadEste, line 26: Excellent use of forEach and arrow function!

checkout.html
```diff
@@ -0,0 +1,177 @@
...
+                            <p class="dog-card-price-description"><span>Cost to Adopt:&nbsp;</span>$123.45</p>
+                        </div>
+                    </div>
+                </article>
+                </section>
+            </div>
+            <div class="form-item">
+                <label>Total Adoption Cost</label>
+                <p>$1234.56</p>
+                <button type="submit" onclick="">Submit</button>
```
MadEste: Not necessary to include onclick if will be empty

checkout.html
```diff
@@ -0,0 +1,177 @@
...
+                <a href="index.html">Home</a>
+                <span>/</span>
+                <a href="dogs.html">Dogs</a>
+                <span>/</span>
+                <a href="blog.html">Blog</a>
+            </nav>
+        </header>
+    <main>
+        <div class="main-heading"><h1>Checkout</h1></div>
+        <form onsubmit="onSubmit();formValidation()">
```
MadEste: Might be better practice to combine both into one function. Also your page reloads before console can be viewed. you can prevent this by returning false in your onSubmit function. 

Combine both into one function and do `<form onsubmit="return onSubmit()">`

Then at the end of your onSubmit function, return false.

scripts/script.js
```diff
@@ -0,0 +1,114 @@
...
+    let formValuesArray = [];
+    form.forEach(e => {
+        formValuesArray.push(e.value);
+        if (e.checked) {
+            console.log(e.value + " is checked");
+        } else {
+            console.log(e.value);
+        }
+    })
+    console.log(formValuesArray);
```
MadEste, line 106: Nice work logging all values correctly.

PR #30, opened 2020-04-03
Set up all required pages of the project

jak-ashuraliev:
Setup the following pages: index.html, dogs.html, blog.html, checkout.html, and styles.css


jak-ashuraliev (commented):
Please disregard this change request d5f6b21. 
I was able to debug the issue.
Thanks.

styles.css
```diff
@@ -0,0 +1,429 @@
...
+.checkout-wrapper {
+  padding-top: 3rem;
+}
+
+header p {
+  color: #8bc34a;
+  font-size: 1.5rem;
+  font-style: italic;
+  position: relative;
+  top: -14px
```
MadEste: Make sure to update px to responsive units: %, em, rem. px are only acceptable for border line width.

index.html
```diff
@@ -0,0 +1,194 @@
...
+      <div class="card">
+        <figure>
+          <img onclick="alertDogInfo('Angus', 'Boxer', 123.45)" src="images/angus-card.jpg" alt="Staring sad dog">
+          <h4>Angus</h4>
+          <small><strong>Cost to Adopt:</strong> $123.45</small>
+          <figcaption>Consectetur tempora magnam optio quibusdam! Lorem ipsum dolor sit amet, consectetur adipisicing
+            elit.
+          </figcaption>
+        </figure>
+        <a href="#" onclick="addToTotal(123.45)" class="btn btn-adopt">Adopt</a>
```
MadEste: I would recommend using a button element instead of anchor so the href="#" does not force the window to the top of the page on each click.

blog.html
```diff
@@ -0,0 +1,122 @@
...
+      </ul>
+    </nav>
+    <!-- NAV END -->
+
+    <!-- SECTION START -->
+    <section class="section-blog">
+
+      <h2>Adoptadog Blog</h2>
+
+      <div class="blog-box">
```
MadEste: You should no longer have these elements in html and only use the script to generate blog posts.

styles.css
jak-ashuraliev: Updated. Thanks.

index.html
jak-ashuraliev: Great Advice - Updated. Thanks.

blog.html
jak-ashuraliev: I'm not sure if I got to that point yet. Assignment requirements didn't ask me to use scripts to generate blogs. I believe it will be covered in jQuery. Thank you.

styles.css
```diff
@@ -0,0 +1,436 @@
...
+}
+
+main {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-between;
+}
+
+.card {
+  border: 1px solid #d3d3d3;
```
MadEste, line 207: px are acceptable here!

styles.css
```diff
@@ -0,0 +1,1016 @@
...
+
+/* #093b5f    dark-blue*/
+/* #8bc34a    green*/
+/* #ebebeb    light-grey*/
+
+/* =============== GLOBAL =============== */
+
+* {
+  margin: 0;
+  padding: 0;
```
MadEste, line 12: Good work removing any user stylesheet margins etc.

styles.css
```diff
@@ -0,0 +1,1016 @@
...
+  height: 25rem;
+  color: #fff;
+  display: flex;
+  justify-content: space-between;
+  align-items: flex-start;
+}
+
+header::before {
+  background: linear-gradient(to top, green, rgb(90, 90, 90) 50%);
+  z-index: -30;
```
MadEste, line 94: Can probably just combine both these header::before{
the z-index on line 106 will overwrite this.

blog.js
```diff
@@ -0,0 +1,59 @@
+const blog = document.querySelector('#blog');
+const blogContent = [{
```
MadEste, line 2: Good work using an array of objects to store information.

blog.js
```diff
@@ -0,0 +1,59 @@
...
+
+
+  blog.appendChild(blogBoxDiv);
+  blogBoxDiv.appendChild(div);
+  div.appendChild(blogImg);
+  blogBoxDiv.appendChild(blogContentDiv);
+  blogContentDiv.appendChild(blogContentTitle);
+  blogContentDiv.appendChild(blogText1);
+  blogContentDiv.appendChild(br);
+  blogContentDiv.appendChild(blogText2);
```
MadEste, line 58: 👍

script.js
```diff
@@ -0,0 +1,40 @@
...
+  e.preventDefault();
+
+  let formValues = {
+    name: document.getElementById('name').value,
+    email: document.getElementById('email').value,
+    address: document.getElementById('address').value,
+    city: document.getElementById('city').value,
+    state: document.getElementById('state').value,
+    zipcode: document.getElementById('zipcode').value,
+    firstAdoptY: document.getElementById('yes').value,
```
MadEste: Be careful with how you treat radio buttons. They have an extra field called checked that you should report instead of value. As it is now your form always reports yes for a first time adopter.

PR #36, opened 2020-04-18
fixed errors

minhochoi1983:


style.css
```diff
@@ -0,0 +1,455 @@
...
+    right: 100px;
+    top: 100px;
+    color: white;
+    display: block;
+    vertical-align: middle;
+    
+}
+
+.cart-img {
+    width: 50px;
```
MadEste: Be careful about using px in the next assignment. px are not responsive units and should not be used ANYWHERE in your css except to perhaps set the line width of an element.

style.css
```diff
@@ -0,0 +1,455 @@
...
+    height: 500px;
+    overflow: hidden;
+    
+}
+
+
+
+.header-left {
+    position: absolute;
+    top: 100px;
```
MadEste: Do not use px for positioning moving forward.

style.css
```diff
@@ -0,0 +1,455 @@
...
+    font-weight: 900;
+    color: #164668;
+}
+
+nav ul li .menu-lists {
+    color: #164668;  
+}
+nav {
+    background-color: #EBEBEB;
+    height: 50px;
```
MadEste: try em, rem, or %

minhochoi1983:
Edited all px to rem

minhochoi1983:
added lesson 4 parts

PR #37, opened 2020-05-18
Updated Adoptadog Site

ksjolly:


blog.html
```diff
@@ -0,0 +1,119 @@
...
+      <li><a href="index.html">Home  </a></li>
+      <li><a href="dogs.html">Dogs</a></li>
+      <li><a href="blog.html">Blog</a></li>
+    </ul>
+  </nav>
+  <div>
+    <section class = "sectionblog">
+      <h1 class="dogsh1"> Adoptadog Blog </h1>
+      <div class="clearfix">
+      <img class="mainimage" src="images/blog-1.jpg" alt="Smiley face" height="35%" width="35%">
```
MadEste: You can assign % values in css, but in html theses should just be numbers and refer to px size.

checkout.html
```diff
@@ -0,0 +1,111 @@
...
+            <label class = "label">City</label>
+            <select id="country" name="country">
+              <option value="australia">Australia</option>
+              <option value="canada">Canada</option>
+              <option value="usa">USA</option>
+            </select>
+
+          <!-- </div>
+          <div class = "form-div"> -->
+            <label class = "label" style = "padding-left: 10px;">State</label>
```
MadEste, line 70: Try not to use inline styles in html. Best practice nowadays is to apply all styles inside the stylesheet. Also px are not a responsive unit. Outside of defining border widths, we recommend using %, em, rem units moving forward.

styles/stylesheet.css
```diff
@@ -0,0 +1,414 @@
...
+
+.divheader {
+
+}
+.button-bottom {
+  width: 20%;
+      background-color: #ebebeb;
+      border: none;
+      color: darkslategrey;
+      padding-top: 15px;
```
MadEste: Styles look good for now, but you will need to update these to responsive units instead of px for the next assignment.

styles/stylesheet.css
```diff
@@ -0,0 +1,414 @@
...
+.mainh1 {
+  margin-top: 0px;
+  font-size: xx-large;
+  font-family: 'Neuton', serif;
+  color: darkslategrey;
+
+}
+
+.dogsh1 {
+  font-size: xxx-large;
```
MadEste, line 126: I would ignore the issue with the validator for now as this should not be creating an error. However I would advise against font size keywords and instead us em or rem for font-size. These keywords can vary from browser to browser and explicitly stating the size will produce more accurate results.

checkout.html
```diff
@@ -0,0 +1,111 @@
...
+            <select id="country" name="country">
+              <option value="australia">Australia</option>
+              <option value="canada">Canada</option>
+              <option value="usa">USA</option>
+            </select>
+
+          </div>
+          <label class = "label"> First Time Adopter? </label>
+          <div style = "margin-top: 10px; margin-bottom: 10px">
+            <input type="radio" id="other" name="gender" value="other"> <span style = "margin-right: 40px"> Yes </span>
```
MadEste: Id's should be unique and only used once per page element. Use the 'name' attribute to group radio buttons together instead. 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio

blog.html
MadEste: I would recommend not using height and width attributes and instead leave it to the html styling for sizing.

PR #41, opened 2020-07-13
Course Project

Steve-Carlton:
Hello!

Here are my assignment submissions. Thank you!  -S


Steve-Carlton:
Hello!

Here is my CSS for assignment 2. Thank you! -S

Steve-Carlton:
Hello! See updates for assignment 3.4 adding basic responsiveness. Thanks! -S

Steve-Carlton:
Hello! Added responsiveness to assignment 3.4. Thank you for the assist! -S

Steve-Carlton:
Hello! JS is linked. Assignment 4.2 complete. Thanks! -S

Steve-Carlton:
Assignment 6 completed. Thanks! -S

style.css
```diff
@@ -0,0 +1,542 @@
...
+#blog-preview p {
+  line-height: 1rem;
+  }
+
+#blog-preview a h6 {
+  margin: 0;
+  }
+
+.current-dogs-container {
+  width: 60%;
```
tylervanbrocklin: Would recommend changing to 80%, with 10% margin on each side it will make your content look more centered. Just a suggestion!

Steve-Carlton: Thank you for the feedback Tyler! I'll give it a go. -S

Steve-Carlton: Hello Tyler, I applied the suggested percentages, but found that a width of 80% on .current-dogs-container made the content spill to the right and encroached on the blog preview section. 63% was the most I could increase it's size comfortably. Am I interpreting your comment correctly?

Steve-Carlton: Thread resolved via email.

Steve-Carlton:
Hello! Assignment 7 submitted.

adoption.html
```diff
@@ -0,0 +1,154 @@
...
+    <div class="title">
+      <a href="index.html"><img src="images/adoptadog-logo.svg" alt="Adoptadog logo" class="logo"></a>
+      <h1>Seattle's premiere dog adoption agency</h1>
+    </div>
+    <div class="cart">
+      <div class="cartline1">
+        <img src="images/cart-icon.svg" alt="your checkout cart">
+        <span>$1234.56</span>
+      </div>
+      <h5 class="cartline2"><a href="checkout.html">Checkout</a></h5>
```
tylervanbrocklin, line 23: Take a look at the CSS for `.cart a {` it appears you're setting `top: 1.5rem`. On mobile part of your navbar is covered by the element, therefore preventing me from clicking. While absolute sizing have valid uses, strange bugs like this can appear when mixing with responsive code.

tylervanbrocklin: On desktop, try pushing your browser to mobile size and click on 'Dogs' in your navbar to see. You can inspect element here and select your <a> tag in the cart to see what is happening.

blog.html
```diff
@@ -0,0 +1,91 @@
...
+
+  <nav class="topnav">
+    <h2>
+      <a href="index.html">Home</a> <span>/</span> <a href="adoption.html">Dogs</a> <span>/</span> <a href="blog.html">Blog</a>
+    </h2>
+  </nav>
+
+  <h3 class="headline">Adoptadog Blog</h3>
+  <section class="intro" id="blog-page">
+    <article class="post">
```
tylervanbrocklin, line 37: I noticed your blog articles use the `overflow` css property. While this can be useful in some cases, it is generally not a good practice to use them unless absolutely necessary. In this case, it seems unnecessary to make your blog articles scroll horizontally when you should be making the content (image, text, etc.) fit the screen properly by making your <article> tags and their content use responsive sizes.

adoption.html
Steve-Carlton: Good catch Tyler, thank you! I switched the top position to a percentage. And I also noticed that I neglected to set a height for the cart in the CSS rules. After setting a height, the cart element no longer covers up the navigation at mobile sizes.

blog.html
Steve-Carlton: Thanks again Tyler! I added a flexbox to the .intro CSS rule. I did notice however that the blog images on the blog page travel down the page as the browser size narrows (most noticeable at a browser width around 871px). As a result the images are misaligned with the corresponding type for each blog post. I'm not sure what is causing this thoughts?

Steve-Carlton: Thread resolved via email

scripts/autoblog.js
```diff
@@ -0,0 +1,143 @@
...
+entryTitle3.textContent = 'Teach Your Dog To Fetch!';
+
+//create blog body copy
+let entryCopy3 = document.createElement('p');
+entryCopy3.textContent = 'Iduciendisite quo magnatem iuntum quid quaest ea am, tenderumet adis dolenem quidustrum fuga. Faceaquae estioria derum recuptatur, cum volore, undipsa doloreium hillupta aut es ut alitatuscit ommossum haritatur arum qui of cae videbiti corporeium faccull oribus es quidignis ipietus explam sus am aut amet ant fugiatum, utem non reptat. Uptiusd andesci qui nem aut vendion ectur? Debis que explaut laborenia que doluptur, con et labor abor sant poreperum dio quat que doluptatur aut voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam.';
+//Blog Entries END
+
+// Blog objects. 1 object per blog post.
+const blogObject1 = {
+  post: post1,
```
tylervanbrocklin, line 103: A simpler method of doing this would be to put the `text` or `img url` directly into your array of objects ( see example below ). Then in your `for()` loop you can create / append the elements directly. This would cut down on the length of your code quite a bit.

```
const blogObjects = [
    { post: "sdfsdfsdfsdf", image: "img/url/here", title: "title"}
    { post: "sdfsdfsdfsdf", image: "img/url/here", title: "title"}
    { post: "sdfsdfsdfsdf", image: "img/url/here", title: "title"}
];
```

scripts/checkout.js
```diff
@@ -0,0 +1,56 @@
...
+  //
+  console.log('Pickup Loc: ' + pickupLoc.value);
+  alert('Thank you. The form information has been received');
+}
+//As of Assignment 9, above code is unused and has been reinterpreted into jQuery. See below.
+
+//Assignment 9: refactor console logging form inputs w/ jQuery.
+
+$('.submit').click(function (e) {
+  // myFunction(event);
```
tylervanbrocklin, line 38: Make sure to include prevent default here. Had to comment out this function for grading because it was triggering the refresh. All good because I see you are using this for A9.

scripts/checkout.js
```diff
@@ -0,0 +1,56 @@
...
+  let custmrState = document.getElementById('custmrState');
+  let custmrZip = document.getElementById('custmrZip');
+  let custmrNewYes = document.getElementById('custmrNewYes');
+  let custmrNewNo = document.getElementById('custmrNewNo');
+  let pickupLoc = document.getElementById('pickupLoc');
+
+  console.log('Name: ' + custmrName.value); //logs values of field.
+  console.log('Contact: ' + custmrContact.value);
+  console.log('Address: ' + custmrSt.value);
+  console.log(custmrCity.value);
```
tylervanbrocklin: If you're going to group address together, I would recommend combining all values onto the same `console.log("address:" + st + ", " + city + ", " + state + " " + zip)`.

scripts/checkout.js
```diff
@@ -0,0 +1,56 @@
...
+  alert('Thank you. The form information has been received');
+}
+//As of Assignment 9, above code is unused and has been reinterpreted into jQuery. See below.
+
+//Assignment 9: refactor console logging form inputs w/ jQuery.
+
+$('.submit').click(function (e) {
+  // myFunction(event);
+  console.log('Name:');
+  console.log($('#custmrName').val());
```
tylervanbrocklin: Would combine these on the same `console.log("Name: " + $('#custmrName').val())`

Steve-Carlton: Great catch! Makes perfect sense! Thank you!

scripts/checkout.js
Steve-Carlton: Oh, definitely. Good call! -S

scripts/checkout.js
Steve-Carlton: Thanks Tyler!

scripts/autoblog.js
Steve-Carlton: Interesting! I think I understand the first part of your comment. So, make an array for each blog post and make an object that contains all arrays? 
With this way, I'm unsure of how to write the array. How would all the classes and attributes get assigned? I've attached a codepen link to an excerpt of my JS if that helps. Thanks! -S

https://codepen.io/SteveCarlton/pen/mdEdpoe?editors=0010

Steve-Carlton:
Assignment 9 submitted. Thanks!

scripts/checkout.js
```diff
@@ -0,0 +1,47 @@
...
+  console.log('Pickup Loc: ' + pickupLoc.value);
+  alert('Thank you. The form information has been received');
+}
+//As of Assignment 9, above code (Assignment 8) is unused and has been reinterpreted into jQuery. See below.
+
+//Assignment 9: refactor console logging form inputs w/ jQuery.
+
+$('.submit').click(function (e) {
+  // myFunction(event);
+  event.preventDefault();
```
tylervanbrocklin: You'll want to make sure to use `e.preventDefault()` or change the parameter from `e` to `event`

scripts/autoblog.js
```diff
@@ -0,0 +1,145 @@
...
+  post: post3,
+  image: img03,
+  copyContainer: blogCopyContainer3,
+  title: entryTitle3,
+  bodyCopy: entryCopy3,
+};
+
+// Blog objects listed in an array (for loop)
+const objArray = [blogObject1, blogObject2, blogObject3];
+
```
tylervanbrocklin, line 142: Since you aren't referencing your blog objects individually. You can actually just create them inside of the array:
```
const blogPosts = [
   {
      // post 1
   },
   {
      // post 2
   },
   {
      // post 3
   }
]
```

scripts/ada-01.js
```diff
@@ -0,0 +1,26 @@
...
+let total = 0;
+// function costCalculator(x) {
+//   total += x;
+// }
+
+//Running cart total refactored in jQuery
+$('.add-to-cart')
+  function costCalculator(x) {
+    total += x;
+    alert('Cart Total: $' + total);
```
tylervanbrocklin, line 25: Glad to see you were able to refactor this one well!

scripts/checkout.js
```diff
@@ -0,0 +1,47 @@
...
+  } else {
+    console.log('New customer: No');
+  }
+  //
+  console.log('Pickup Loc: ' + pickupLoc.value);
+  alert('Thank you. The form information has been received');
+}
+//As of Assignment 9, above code (Assignment 8) is unused and has been reinterpreted into jQuery. See below.
+
+//Assignment 9: refactor console logging form inputs w/ jQuery.
```
tylervanbrocklin, line 35: Your comments make understanding your code very easy! Especially with the last few assignments overlapping in some places.

scripts/autoblog.js
Steve-Carlton: Cool, that hadn't occurred to me, thanks! Would this create neater code or are there potential performance gains with this way?

scripts/checkout.js
Steve-Carlton: Hm, I don't yet understand differences between e or event enough to fully appreciate your comment. I'll look have to look into it. Thanks!

scripts/ada-01.js
Steve-Carlton: Cool, thanks for the assist!

scripts/checkout.js
Steve-Carlton: Thanks! They make it easier for me as well! I'm glad to hear they're useful. I was wondering if it would too cluttered.

PR #53, opened 2020-12-05
Lesson6

NayelaAhmed:


html/styles.css
```diff
@@ -0,0 +1,477 @@
...
+
+  .introduction .intro-image {
+    width: 100%;
+    margin-right: 40px;
+    display: block;
+  }
+}
+
+.dog-grid, .all-dogs-grid {
+  display: grid;
```
tylervanbrocklin, line 199: Did you mean to include these in the media query above it?

html/styles.css
```diff
@@ -0,0 +1,477 @@
...
+  flex-direction: row-reverse;
+}
+
+.home-section {
+  flex-shrink: 2;
+}
+
+.blog-preview {
+  flex-basis: 450px;
+  display: none;
```
tylervanbrocklin, line 108: We want the blog to show on iPad and Desktop views. Right now it only seems to show on desktop. I think this is from the @media query below. You'll want to use `display: block` on min-width of iPad. Therefore it'll show on iPad and Desktop. Then on your mobile media query you'd use `display: none`

html/styles.css
```diff
@@ -0,0 +1,477 @@
...
+}
+
+@media only screen and (max-width:500px) {
+  .introduction {
+    flex-direction: column;
+    align-items: flex-start;
+  
+  }
+
+  .introduction .intro-image {
```
tylervanbrocklin, line 191: Right now, when adjusting my screen size I noticed the into-image becomes tiny under tablet size. Using width percentage takes in from it's parent. Would look at using a responsive size like rem

html/styles.css
```diff
@@ -0,0 +1,477 @@
...
+  grid-template-columns: auto auto auto auto;
+}
+
+@media only screen and (max-width: 768px) {
+  .dog-grid {
+    grid-template-columns: auto auto; 
+  }
+
+  .all-dogs-grid {
+    grid-template-columns: auto auto auto;
```
tylervanbrocklin, line 216: You will want the same number of columns for the home page and the dog page

NayelaAhmed: I thought the dog page is suppose to have 4 columns when I looked at the mockup. This is why I added a second class name to the home page grid so they behave differently.

html/styles.css
NayelaAhmed: I made the min-width media query too big instead of 768px it says 1200px. Must have got mixed up with different branches.

html/styles.css
NayelaAhmed: No, I took care of the dog grid below. 
@media only screen and (max-width: 768px) {
  .dog-grid {
    grid-template-columns: auto auto; 
  }

  .all-dogs-grid {
    grid-template-columns: auto auto auto;
  }
}

html/styles.css
NayelaAhmed: This is how it appears for me. It looked normal, but I will look into it.

homepg-snippet

PR #55, opened 2020-12-12
Lesson07

NayelaAhmed:


html/index.html
```diff
@@ -0,0 +1,201 @@
...
+
+                </div> <!-- end of blog-preview-snippet  -->
+
+
+            </aside>
+            
+
+            <section class="home-section">
+                <div class="introduction">
+                    <img class="intro-image"
```
tylervanbrocklin, line 96: Your intro image does not scale properly. It looks tiny on desktop, then enlarges on mobile.

html/styles.css
```diff
@@ -0,0 +1,506 @@
...
+body {
+  font-family: Arial, Helvetica, sans-serif, serif;
+}
+
+.page-header {
+  position: relative;
+  display: block;
+  height: 400px;
+  overflow: hidden;
+  background-image: url('../images/header-image.jpg');
```
tylervanbrocklin, line 18: You can remove this line, it is being repeated in the line below

PR #62, opened 2021-01-17
Meghan Cahill's Assignment 1

Cahillmn:
I added the html and some styling to the adopt a dog website


styles/style.css
```diff
@@ -0,0 +1,382 @@
+* {
+    font-size: 1.15rem;
```
jimmynono: another good thing to include in here is `box-sizing: border-box;` I'll go over it in class.

jimmynono: a good strategy for font sizes is to use this.

```html {
  font-size: 100%;
}```

This sets the base font at 100% of the user 's settings (default is 16px) after that you should set the individual elements lke 
```
h1 {
  font-size: 2.5rem
}

h2 {
  font-size: 2rem;
}
```
etc

File: styles/style.css [hunk 2]
```diff
@@ -0,0 +1,382 @@
...
+
+button {
+    text-align: center;
+    color: white;
+    background-color: rgb(18, 146, 18);
+    font-weight: bold;
+}
+
+.dogs button {
+    margin-left: 10%;
```
jimmynono: generally you'll want the same padding each side  horizontally and the same padidng vertically; 

you can shorthand this (using your values...) to
`margin: 7% 10%` which will do 7% top/bottom and 10% left/right.

I'd also suggest  using REM's for padding/margin

styles/style.css
```diff
@@ -0,0 +1,382 @@
...
+    /* margin-right: 10%;  */
+    margin-bottom: 3%;
+}
+
+.city-state input, .city-state select {
+    margin-top: 5%;
+
+} 
+
+#zipcode {
```
jimmynono: use classes for CSS. Even if it's just used once, that's ok.

styles/style.css
```diff
@@ -0,0 +1,378 @@
...
+    grid-template-rows: auto auto;
+    grid-column-gap: 3%;
+    grid-row-gap: 5%;
+    margin-left: 15rem;
+    padding-bottom: 3rem;
+}
+
+.blog {
+    display: grid;
+    grid-template-columns: 30% 40%;
```
tylervanbrocklin: I would advise using flexbox over CSS grid for most things. In the case of your dog grid, you can go either way. In this case, you'll find using flexbox will simplify your code and make your site much more responsive. Here is a cheat sheet I always use for referencing flexbox: https://css-tricks.com/snippets/css/a-guide-to-flexbox/

index.html
```diff
@@ -0,0 +1,164 @@
...
+                arum qui officae videbiti corporeium faccull oribus es quidignis ipietus explam sus
+                am aut amet ant fugiatum, utem non reptat ad magnienecto volorent fugia nonsedi gendand itatquos as
+                magnimpor aut molor autatia ad est officimus accae et omnia cus, od qui ate</p>
+        </div>
+
+    </section>
+
+    <section class="dogs">
+
+        <article class="row-1 col-1">
```
tylervanbrocklin: When shrinking down screen size, your dogs become too small to read. I would suggest utilizing minmax() and autofit to create flexible column sizes. This allows you to create a grid that responds well to any screen size! Check out this article on it here: https://mastery.games/post/tile-layouts/#minmax-Has-your-Back

You will notice this method will also eliminate the need for your row/col classes (row-1, col-1, row-2, col-2...) classes.

index.html
```diff
@@ -0,0 +1,164 @@
...
+
+        </ul>
+    </nav>
+
+    <section class="adopt">
+
+        <img class="col-1 row-1" src="images/home-intro.jpg" alt="Man hugging a dog">
+
+        <div class="col-2 row-1">
+            <h2>Adopt A Dog</h2>
```
tylervanbrocklin, line 45: I would use flexbox for this over css grid. When changing screen sizes, your image gets tiny and text pressed against the side. Using flexbox, you can leverage `flex-wrap: wrap` to make your text go under your image on a smaller screen size, more details here: https://css-tricks.com/snippets/css/a-guide-to-flexbox/

js/dogs.js
```diff
@@ -0,0 +1,10 @@
+function pictureInfo(dog, breed, price) {
+    result = 'Name: ' + dog + ' Breed: ' + breed + ' Price: ' + price;
```
tylervanbrocklin: All new variables must be declared with either let or const.

styles/style.css
```diff
@@ -0,0 +1,413 @@
...
+        width: 100%;
+        grid-template-columns: repeat(auto-fill, minmax(0rem, 100%));
+        overflow: auto;
+    }
+}
+
+@media screen and (min-width: 576px) and (max-width: 1279px) {
+    .dogs {
+        width: 100%;
+        grid-template-columns: repeat(auto-fill, minmax(0rem, 45%));
```
tylervanbrocklin, line 210: good work adjusting the grid columns with media queries!

styles/style.css
```diff
@@ -0,0 +1,413 @@
...
+}
+
+.blog-preview p {
+    margin-top: 1rem;
+    margin-bottom: 1rem;
+}
+
+@media screen and (max-width: 575px) {
+    .blog-preview {
+        display: none;
```
tylervanbrocklin, line 240: Nicely done getting the blog preview to disappear on mobile

js/blog.js
```diff
@@ -0,0 +1,53 @@
...
+    {
+        'image': 'images/blog-3.jpg',
+        'alt': 'Person playing ball with their dog in the sunset',
+        'h3': 'Teach Your Dog To Fetch!',
+        'p': 'Iduciendisite quo magnatem iuntum quid quaest ea am, tenderumet adis dolenem quidustrum fuga. Faceaquae estioria derum recuptatur, cum volore, undipsa doloreium hillupta aut es ut alitatuscit ommossum haritatur arum qui officae videbiti corporeium faccull oribus es quidignis ipietus explam',
+        'p2': 'Uptiusd andesci qui nem aut vendion ectur? Debis que explaut laborenia que doluptur, con et labor abor sant poreperum dio quat que doluptatur aut voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam.'
+    }
+]
+
+for (let i = 0; i < blogInfo.length; i++) {
```
tylervanbrocklin, line 28: Great work using the for loop to generate your blog posts!!

styles/style.css
```diff
@@ -0,0 +1,420 @@
...
+    flex-direction: row;
+    justify-content: space-between;
+    align-content: space-between;
+    margin-bottom: 3rem;
+
+}
+
+.blog img {
+    width: 100%;
+    height: 300px;
```
tylervanbrocklin, line 256: This will make your blog images stretched on mobile. I would suggest setting a width, such as 20rem. Then set your height to auto. `height: auto;`

js/dogs-jquery.js
```diff
@@ -0,0 +1,13 @@
+$('.grey-border').on('mouseover', function (e) {
```
jimmynono, line 1: One of the cool things about this, is you may not need to remove the grey-border class as the class you add in on hover might override it. But a good thought process.

PR #63, opened 2021-01-18
Sean Estes Pull Request for Lesson 01

estessm:
This is my HTML and some CSS for lesson 01.  Looking forward to input.

Thanks,
Sean


estessm:
Updated code for lesson 2 updates

styles/styles.css
```diff
@@ -0,0 +1,348 @@
...
+header {
+  background-color: #0A3C61;
+  height: 24rem;
+  justify-content: flex-start;
+
+}
+
+header h1 {
+  color: black;
+  position: absolute;
```
jimmynono, line 24: I'd argue that positioning these using position might not be best. See what you can do with padding/margin

styles/styles.css
```diff
@@ -0,0 +1,348 @@
...
+  left: 13rem;
+  top: 1rem;
+  font-size: 60px;
+  font-family: sans-serif;
+}
+
+header h2 {
+  color: white;
+  position: absolute;
+  font-size: 25px;
```
jimmynono: Make sure to use rem's for font sizes. You can just do math 25px / 16 = 1.5625rem (you could round down to 1.5 rem if you'd like.

styles/styles.css
```diff
@@ -0,0 +1,348 @@
...
+  font-family: sans-serif;
+  font-style: italic;
+  line-height: 36px;
```
jimmynono: Rem's here are also appropriate

styles/styles.css
```diff
@@ -0,0 +1,348 @@
...
+}
+
+.b-box {
+  width: 30rem;
+  height: 100px;
```
jimmynono: make sure you're consitent with px/rem on heights. Below you use REM's here you use px.

Either is fine, but I prefer REMS

styles/styles.css
```diff
@@ -0,0 +1,348 @@
...
+
+
+
+/* Style the links inside the navigation bar */
+nav {
+  font-family: sans-serif;
+}
+
+nav li {
+  padding: 0 5px 0 5px;
```
jimmynono: You can shorthand this to `padding: 0 5px;`

styles/styles.css
```diff
@@ -0,0 +1,348 @@
...
+  position: relative;
+  left: 11rem;
+}
+
+nav li a:after{
+  content: " / ";
```
jimmynono, line 114: Woot!

styles/styles.css
```diff
@@ -0,0 +1,348 @@
...
+  position: absolute;
+  right: 0;
+  top : 15px;
+}
+
+nav li:last-child a:after{
```
jimmynono, line 120: If you want to challenge yourself with these pseudo elements. See how you could leverage :not

styles/styles.css
```diff
@@ -0,0 +1,348 @@
...
+  list-style-type: none;
+  margin: 0;
+  padding: 0;
+  height: 50px;
+  overflow: hidden;
+  background-color: rgb(235, 235, 235);
+}
+
+nav li {
+  float: left;
```
jimmynono: Don't use floats for this. probably `display:inline-block;` would work

styles/styles.css
```diff
@@ -0,0 +1,348 @@
...
+  padding: 14px 16px;
+  text-decoration: none;
+}
+
+/* Change the link color to #111 (black) on hover */
+nav li a:hover {
+  background-color: Grey;
+}
+
+#wrapper {
```
jimmynono: Don't use id's for CSS only use classes. It's ok if it's only used once.

styles/styles.css
```diff
@@ -0,0 +1,348 @@
...
+  font-size: 30px;
+}
+
+.dogimg {
+  width: 100%;
+}
+
+.div1 {
+  width: 98%;
+  display: inline-table;
```
jimmynono: Avoid using table. Tables are really meant for tables of data. Think of using grid or flexbox on laying these out

styles/styles.css
```diff
@@ -0,0 +1,348 @@
...
+
+.div2 {
+  width: 22%;
+  min-width: 275px;
+}
+
+.col {
+    width: 90%;
+    margin: 2rem auto;
+    float: left;
```
jimmynono: No floats. Find a different way. Grid or flexbox is more modern and better

styles/styles.css
```diff
@@ -0,0 +1,348 @@
...
+    padding: 5px;
+}
+
+.row::after {
+    content: "";
+    clear: both;
```
jimmynono, line 266: When you take out float, you can also kill this clearfix

styles/styles.css
```diff
@@ -0,0 +1,348 @@
...
+
+
+
+#a1 {
+  font-size: 24px;
+  font-family: sans-serif;
+  color: rgb(44, 154, 183);
+}
+
+#a2 {
```
jimmynono: no id's pls

styles/styles.css
```diff
@@ -0,0 +1,348 @@
...
+  color: rgb(44, 154, 183);
+}
+
+em {
+  font-size: 25px;
+}
+
+
+.body-link {
+  color: rgb(44, 154, 183);
```
jimmynono: Be consistent with colors. I'm seeing the css names (hotpink), hex values, and rgb vales. Mxing css names and one other option is fine, but be consistent. I'd steer you to the hex values  like`#e4e4e4` and a few CSS names if you'd like (like hotpink)

RGB's are great for the cases where you need opacity, or things like box shadows

styles/styles.css
estessm: updated to class

styles/styles.css
estessm: replaces with rem

styles/styles.css
estessm: replaced with rem

styles/styles.css
estessm: replaced with rem

styles/styles.css
estessm: updated for shorthand form

styles/styles.css
estessm: Woot Woot!

styles/styles.css
estessm: got rid of a2 as its not used and updated a1

styles/styles.css
estessm: updated

styles/styles.css
estessm: updated to grid

styles/styles.css
estessm: got rid of float

styles/styles.css
estessm: updated to hex

estessm:
keep the comments coming!  I like learning!

dogs.html
```diff
@@ -0,0 +1,191 @@
...
+        </li>
+        <li>
+          <a href="./blog.html">Blog</a>
+        </li>
+      </ul>
+    </nav>
+
+    <main class="wrapper">
+      <section class="flex-container">
+        <h2 class="div1">Here are all of our Dogs!</h2>
```
tylervanbrocklin, line 41: Awesome work making your dog grid very responsive

dogs.html
```diff
@@ -0,0 +1,191 @@
...
+          <article class="div2">
```
tylervanbrocklin, line 42: Would suggest renaming "div2" to something that better describes what it is used for. Such as "dog-card". When reading through a stylesheet, this makes it much easier to tell what each class does.

dogs.html
```diff
@@ -0,0 +1,191 @@
...
+
+            <img class="dogimg" src="./images/murphy-card.jpg" alt="Brown and White dog with black bakcground lovingly looking at camera">
+
```
tylervanbrocklin, line 45: Would also suggest seeing if you can add some background/border to create the tile/card look!

blog.html
```diff
@@ -0,0 +1,85 @@
...
+        <li>
+          <a href="#">Blog</a>
+        </li>
+      </ul>
+    </nav>
+
+    <main class="grid main-content wrapper">
+      <section>
+        <h2>Adoptadog Blog</h2>
+          <article class="row instructors c-box">
```
tylervanbrocklin: Would suggest using flexbox here to allow wrapping on mobile. Then grouping your h2 & p tag into a separate div. This will allow you to have the text aligned on the right of each blog image.

checkout.html
```diff
@@ -0,0 +1,164 @@
...
+          <a href="./dogs.html">Dogs</a>
+        </li>
+        <li>
+          <a href="./blog.html">Blog</a>
+        </li>
+      </ul>
+    </nav>
+
+    <main class="wrapper">
+      <section class="left, liability">
```
tylervanbrocklin: Great work here, this page is very responsive

app.js
```diff
@@ -0,0 +1,12 @@
+
+function info(name, breed, price) {
+  alert(name + ',' + ' ' + breed + ',' + ' ' + price);
```
tylervanbrocklin, line 3: Would suggest: `alert(name + ', ' + breed + ', ' + price);` you can include the space with the comma. Also, a easier way is using string interpolation: `alert('Name: ${name} Breed: ${breed} Price: ${price}')` which allows you to do it all without the plus signs. Either works though!!

app.js
```diff
@@ -0,0 +1,12 @@
+
+function info(name, breed, price) {
+  alert(name + ',' + ' ' + breed + ',' + ' ' + price);
+  }
+
+
+let total = 0;
+
+function carttot(price) {
```
tylervanbrocklin, line 9: Small thing, but we generally use camel casing with names. So you'd want to do `function cartTot()`

styles/styles.css
```diff
@@ -0,0 +1,389 @@
...
+  text-decoration: none;
+}
+
+/* Change the link color to #111 (black) on hover */
+nav li a:hover {
+  background-color: Grey;
+}
+
+.wrapper {
+  width: 80%;
```
jimmynono, line 151: This is a good idea at wide widths, but not at narrow widths. A better way to handle this is to do width: 100% and combine with a max width to keep it from getting too big.

styles/styles.css
```diff
@@ -0,0 +1,389 @@
...
+@media screen and (min-width: 768px){
+  .flex-container{
+    display: flex;
+    width: 75%;
+  }
+}
+
+
+
+.flex-container3 {
```
jimmynono, line 210: This class name is hard to understand. It's really important to use a semantic name for these kinds of things. `.blog-preview` or `.sidebar` or something

styles/styles.css
```diff
@@ -0,0 +1,389 @@
...
+  border: 4px solid #A9A9A9;
+  display: inline-block;
+  margin: 5px 5px;
+  padding: 1rem;
+}
+
+.div3 {
+  width: 52%;
+  height: auto;
+  min-width: 275px;
```
jimmynono, line 249: This doesn't quite work. The min-width will overflow. I'd take that off and have teh width be at 100%. Maybe increase the width of your sidebar from 25% to 33% or somethign to keep it from getting too narrow.

index.html
```diff
@@ -0,0 +1,204 @@
...
+        <li>
+          <a href="./blog.html">Blog</a>
+        </li>
+      </ul>
+    </nav>
+
+    <main class="wrapper">
+      <section class="flex-container">
+        <h2 class="hide">Section Header</h2>
+          <article class="div1">
```
jimmynono, line 43: These articles need to all have the same class. You use div1 and div2.

styles/styles.css
```diff
@@ -0,0 +1,389 @@
...
+
+.flex-container {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-around;
+  width: 59%;
+}
+
+@media screen and (min-width: 768px){
+  .flex-container{
```
jimmynono, line 202: This is squishing too much. Instead keep the container width at 100% and adjust the size of the dog tiles

blog.js
```diff
@@ -0,0 +1,53 @@
...
+    text: "mply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum.",
+    image: './images/blog-3.jpg',
+    alt: 'woman playing with dog in sunset',
+  },
+];
+
+const blogsContainer = document.getElementById('blogs-container');
+console.log(blogsContainer);
+
+for (let i = 0; i < blogs.length; i++) {
```
tylervanbrocklin, line 25: Excellent work dynamically generating your blog posts!!

checkout.js
```diff
@@ -0,0 +1,24 @@
+const form = document.getElementById('checkoutForm');
+
+form.addEventListener('submit', function(event) {
+  event.preventDefault();
+  alert("Thank you. The form information has been received.");
+
+  const formData = document.getElementById('checkoutForm').elements;
+
+  for (let i = 0; i < formData.length - 1; i++) {
+    const val = document.getElementById('checkoutForm').elements[i].value;
```
tylervanbrocklin, line 10: Nice work using a loop to print out all of these values

PR #68, opened 2021-01-20
Liz's Adopt A Dog (HTML & CSS)

AKliz:


css/styles.css
```diff
@@ -0,0 +1,343 @@
+/**Global Colors & Fonts**/
```
jimmynono, line 1: YESS!!! Love the variables.

css/styles.css
```diff
@@ -0,0 +1,343 @@
...
+.row{
+  margin-bottom: 2rem;
+}
+main{
+  max-width: 68.75rem;
+  margin: auto;
+  padding-left:0.75rem;
+  padding-right:0.75rem;
+}
+.d-flex{
```
jimmynono: If you import bootstrap you won't need to do this. It's part of bootstrap.

css/styles.css
```diff
@@ -0,0 +1,343 @@
...
+
+
+/**Header Section**/
+header{
+  color: var(--color-white);
+}
+header{
+background-image: url(../images/header-image.jpg);
+background-size: cover;
+}
```
jimmynono: Make sure and spread out your css a bit.  A space after the selector and a line in between selectors.

image

css/styles.css
```diff
@@ -0,0 +1,343 @@
...
+header nav ul, footer nav ul{
+  text-align: left;
+  padding-top: 1rem;
+  list-style-type: none;
+  overflow: hidden;
+  margin-bottom: 0rem;
+}
+
+nav ul li{
+  float:left;
```
jimmynono: You shouldn't need floats to lay out these elements. `display:inline-block` is a better option

css/styles.css
```diff
@@ -0,0 +1,343 @@
...
+  font-weight: bold
+}
+nav ul li::after{
+    content: "/";
```
jimmynono: Yes!

css/styles.css
```diff
@@ -0,0 +1,343 @@
...
+  text-align: center;
+  padding-top: 4rem;
+  padding-bottom: 4rem;
+}
+.footer-container{
+  max-width: 30%;
+  margin:auto;
+}
+
+@media screen and (max-width:898px){
```
jimmynono: watch your indenting

checkout.html
```diff
@@ -0,0 +1,156 @@
...
+        <nav class="top ">
+          <ul>
+            <li><a href="./index.html" class="active">Home</a></li>
+            <li><a href="./dogs.html">Dogs</a></li>
+            <li><a href="./blog.html">Blog</a></li>
+          </ul>
+        </nav>
+      </div>
+    </header>
+    <main>
```
tylervanbrocklin, line 35: Would suggest adding some padding or margin. Your content sits right against the left side of the screen

blog.html
```diff
@@ -0,0 +1,94 @@
...
+          <ul>
+            <li><a href="./index.html" class="active">Home</a></li>
+            <li><a href="./dogs.html">Dogs</a></li>
+            <li><a href="./blog.html">Blog</a></li>
+          </ul>
+        </nav>
+      </div>
+    </header>
+    <main>
+      <section>
```
tylervanbrocklin: I would suggest adding some padding or margin here to keep your content more to the center

js/javascript.js
```diff
@@ -0,0 +1,12 @@
+//create click handlers that alerts the dogs information when clickin on the dog's photo"
+function dogInfo(dogName, breed, fees){
+  alert(`To adopt ${dogName}, the ${breed}, it will cost ${fees}.`);
```
tylervanbrocklin, line 3: nice job using string concatenation!

css/styles.css
```diff
@@ -0,0 +1,410 @@
...
+  font-family: var(--title-font);
+}
+
+nav a, header nav a {
+  color: var(--color-navy);
+  font-weight: bold;
+}
+
+/**Header Section**/
+header {
```
tylervanbrocklin, line 72: Love how organized your code is

css/structure.css
```diff
@@ -0,0 +1,200 @@
...
+}
+
+/**Columns - SMALL**/
+@media screen and (min-width:500px) {
+  .hidden-xs {
+    display: block;
+  }
+
+  .hidden-sm {
+    display: none;
```
tylervanbrocklin, line 103: Good work getting your columns to scale and hiding your blog on mobile screens

css/styles.css
AKliz: I ended up removing bootstrap and trying to code it myself to practice.

js/form.js
```diff
@@ -0,0 +1,25 @@
...
+  alert('Thank you. The form information has been received.');
+
+  //store data in an array or object
+  const checkoutFormData = form.elements;
+  console.log(checkoutFormData);
+  const submitData = {};
+
+  for(i = 0; i < checkoutFormData.length-1; i++){
+    const fieldData = form.element[i].value;
+    const fieldName = form[i].id;
```
tylervanbrocklin: You had an excellent question on why this works. In JS, there are two ways we typically access properties and values. The first is by using a dot. So if I want the ID of something I can do `something.id`. When it comes to an array of something, we use index. So when you access form[1], that is accessing the 2nd element with the form's array of elements. The form itself is made up of various elements as we can see above when you printed the elements out. So form.element is not an array. Therefore you can't access an index of an array that doesn't exist. Checkout a JSFiddle I made demonstrating this: [https://jsfiddle.net/a5Lhgocs/1/](https://jsfiddle.net/a5Lhgocs/1/)

js/dog-tile.js
```diff
@@ -0,0 +1,42 @@
...
+$('article.dog-card').hover(
+  function(e){
+    $(this).addClass('darkborder');
+  }, function(){
+    $(this).removeClass('darkborder')
+  }
+);
+
+
+//add ripple effect to your button - I'm not the genius of this... was able to put this together with things I found online.
```
jimmynono, line 16: Ha! Welcome to becoming a professional web developer. 😸

PR #69, opened 2021-01-20
Adoptadog project

elingv:


blog.html
```diff
@@ -0,0 +1,81 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+  <head>
+    <meta charset="utf-8">
```
jimmynono, line 4: You will need the meta tag that we referenced in week 1. Make sure it's on all of your HTML pages.

blog.html
```diff
@@ -0,0 +1,81 @@
...
+    <title>Blog</title>
+    <link rel="stylesheet" href="./styles/master.css">
+    <style>
```
jimmynono: make sure to have this style in your css file.

blog.html
```diff
@@ -0,0 +1,81 @@
...
+      body {
+         background-color: lightblue;
+      }
+    </style>
+  </head>
+
+  <body>
+    <header>
+      <img src="images/adoptadog-logo.svg" width="200" alt="adoptadog logo">
```
jimmynono: Don't use any styling attributes in your HTML it should all be in your CSS file.

blog.html
```diff
@@ -0,0 +1,81 @@
...
+          <li><a href="dogs.html">Dogs</a></li>
+          <li><a href="blog.html">Blog</a></li>
+        </ul>
+      </nav>
+    </header>
+
+
+    <main>
+      <h2>Adoptadog Blog</h2>
+        <section class="blog1">
```
jimmynono: Since all of these blogs are styled the same, you could use the same class for each blog section. Also, consider another HTML5 element for these blog _articles_

checkout.html
```diff
@@ -0,0 +1,101 @@
...
+        <label for="street">Street Address</label><br>
+        <input type="text" id="street" name="street"><br>
+
+        <label for="city">City</label><br>
+        <input type="text" id="city" name="city"><br>
+
+        <label for="state">State</label><br>
+        <input type="text" id="state" name="state"><br>
+
+        <label for="zip_code">Zip Code</label><br>
```
jimmynono: nice use of for's and id's. These <br>'s should be necessary. You can use display properties (block, inline-block, inline, etc) to format these out.

dogs.html
```diff
@@ -0,0 +1,178 @@
...
+          <li><a href="dogs.html">Dogs</a></li>
+          <li><a href="blog.html">Blog</a></li>
+        </ul>
+      </nav>
+    </header>
+
+
+    <main>
+      <h2>Here are all of our Dogs!</h2>
+        <section>
```
jimmynono: Great! This is the good method for layout. Nice use of the same 'card' property on all the cards

PR #70, opened 2021-01-20
Joann Cahill Adopt a Dog Pull Request

jcahill53:
Initial commit of HTML with some CSS for assignment 1


styles/stylesheet.css
```diff
@@ -0,0 +1,378 @@
+* {
+    margin: 0;
```
jimmynono, line 2: Another thing good to add in this block is the `box-sizing: border box;` Google it to see how it works.

styles/stylesheet.css
```diff
@@ -0,0 +1,378 @@
...
+
+.blog1 h2,
+.blog2 h2,
+.blog3 h2,
+.blog1 p,
+.blog2 p,
+.blog3 p {
+    margin-top: 0;
+    margin-bottom: 2%;
+}
```
jimmynono, line 52: Each blog should have a common class, that way you won't have to use .blog2, .blog3, etc....just a `.blog` should select all of them. Think of how you could adjust your html to use this.

```
.blog p,
.blog h2 {
    margin-top: 0;
    margin-bottom: 2%;
}```

File: styles/stylesheet.css [hunk 3]
```diff
@@ -0,0 +1,378 @@
...
+    flex-direction: row;
+    justify-content: space-between;
+    align-content: center;
+    font-size: .8rem;
+}
+
+header h1 {
+    color: #008A00;
+    font-style: italic;
+    font-size: 1.25rem;
```
jimmynono, line 106: REMS! Yay!

styles/stylesheet.css
```diff
@@ -0,0 +1,378 @@
...
+
+.hdr-right {
+    margin-right: 10%;
+    margin-top: 5%;
+    margin-bottom: 5%;
+    /* width:5%; */
+}
+
+header button {
+    padding: 5% 40%;
```
jimmynono: Padding with buttons might be better handled with REMS.

styles/stylesheet.css
```diff
@@ -0,0 +1,378 @@
...
+}
+
+header img,
+header p,
+header button {
+    margin-top: 10%;
+}
+
+.cart img {
+    width: 50%;
```
jimmynono: percentage! Yay!

styles/stylesheet.css
```diff
@@ -0,0 +1,378 @@
...
+    font-weight: bold;
+}
+
+nav a,
+footer a {
+    text-decoration: none;
+}
+
+
+#wrapper {
```
jimmynono: for CSS the convention is to use classes only. Even if that class is only used once. Change this to a class. 

You'll use id's (and classes) for JS

styles/stylesheet.css
```diff
@@ -0,0 +1,378 @@
...
+    display: grid;
+    grid-template-columns: 40% 40%;
+    grid-template-rows: auto auto;
+    margin-top: 5%;
+    margin-bottom: 5%;
+}
+
+.dogs,
+.all-dogs {
+    display: grid;
```
jimmynono, line 283: Great! 

We'll learn how to make these different at different widths.

styles/stylesheet.css
```diff
@@ -0,0 +1,378 @@
...
+    grid-template-columns: 25% 25% 25% 25%;
+    grid-template-rows: auto auto;
+    margin-top: 5%;
+    margin-bottom: 15%;
+    grid-gap: 3% 3%;
```
jimmynono: for these I like REM's

styles/stylesheet.css
```diff
@@ -0,0 +1,378 @@
...
+    text-decoration: none;
+    padding: 0 15%;
+    width: 100%;
+
+    color: rgb(20, 66, 109);
+    font-weight: bold;
+
+}
+
+.position-button {
```
jimmynono: This block of HTML is a little nested heavy. Just commenting here on the CSS. 

See if you can't eliminate the position relative and the nested stuff and just use margin to center things. I might better explain in person. When we get to Q&A part of class I might be able to talk through it. Just ask me, as I'll probably forget.

styles/stylesheet.css
```diff
@@ -0,0 +1,422 @@
...
+
+.wrapper,
+.blogs {
+    margin-left: 5rem;
+    /* margin-right: 5%; */
+    /* background-color: lightblue; */
+}
+
+.intro {
+    display: grid;
```
tylervanbrocklin: Right now, when I shrink your page down. Your paragraph text overlaps the image. I would suggest only using CSS grid for your dog grid. Then use flexbox for cases like this. You'll find it will simplify your code here and make this part of the site super responsive.

dogs.html
```diff
@@ -0,0 +1,274 @@
...
+            <li> <a href="./dogs.html">Dogs</a></li>
+            <li>/</li>
+            <li> <a href="./blog.html">Blog</a></li>
+
+        </ul>
+    </nav>
+    <div class="wrapper">
+        <h2>Here are all of our Dogs!</h2>
+
+        <section class="all-dogs">
```
tylervanbrocklin, line 45: Right now your dog tiles seem to spill off of the page. You're definitely on the right track here with CSS grid though. Check out this article and see if this helps. Using some shortcuts, you shouldn't need to actually create col-1 row-1 CSS classes. I think this might also solve the grid overflowing issue, and make your code more responsive: https://css-tricks.com/look-ma-no-media-queries-responsive-layouts-using-css-grid/#layout-2-full-width-image-gallery

dogs.html
```diff
@@ -0,0 +1,274 @@
...
+                <p>
+                    Nullam at urna convallis, ornare turpis ut, fermentum odio. Sed auctor non justo ut imperdiet.
+                    Maecenas
+                    feugiat facilisis mi nec vehicula.
+                </p>
+                <button type="submit" value="Adopt" name="adopt" onclick="sumCost(123.45)">Adopt</button>
+        </article>
+
+        <article class = " col-2 row-1">
+                    <img src="images/poppy-card.jpg" alt="Gray and white Shih-tzu named Poppy"
```
tylervanbrocklin: Your indenting makes it a bit hard to read. Most code editors will have a shortcut to ident for you, but there is also websites like this one you can run your code through to make it easier to read:  https://www.freeformatter.com/html-formatter.html

Doesn't matter whether you use tabs or spaces, but is good practice to indent your code and keep whichever you choose consistent across your code

styles/stylesheet.css
```diff
@@ -0,0 +1,543 @@
...
+    .dogs,
+    .all-dogs {
+        grid-template-columns: repeat(auto-fit, minmax(0rem, 100%));
+    }
+}
+
+@media screen and (min-width: 576px) and (max-width: 1279px) {
+    .dogs,
+    .all-dogs {
+        grid-template-columns: repeat(auto-fit, minmax(0rem, 45%));
```
tylervanbrocklin, line 323: Great work with the media queries. Keep in mind you can have multiple css queries within a media query.

styles/stylesheet.css
```diff
@@ -0,0 +1,543 @@
...
+}
+
+.dogs article {
+    border: lightgray solid 1px;
+}
+
+@media screen and (max-width: 575px) {
+    .dogs,
+    .all-dogs {
+        grid-template-columns: repeat(auto-fit, minmax(0rem, 100%));
```
tylervanbrocklin, line 315: Your grid resizes well!

styles/stylesheet.css
```diff
@@ -0,0 +1,543 @@
...
+}
+
+@media screen and (max-width: 575px) {
+    .main-content {
+        width: 100%;
+    }
+}
+
+.blog-link {
+    display: none;
```
tylervanbrocklin, line 378: good work getting your blog preview to disappear on small screen sizes

styles/stylesheet.css
```diff
@@ -0,0 +1,543 @@
...
+}
+
+.dogs,
+.all-dogs {
+    display: grid;
+    grid-template-columns: repeat(auto-fit, minmax(0, 32%));
+    grid-gap: 1% 1%;
+    margin-top: 1rem;
+    margin-bottom: 3rem;
+    overflow: auto;
```
jimmynono: This cuts off the bottom of your last row of dogs. I'd go ahead and remove it. Auto can behave strangely some times.

jcahill53: I used overflow:auto to create the scroll because when I view my dog pages on an iPad and iPhone the dog pictures are overflowing onto my button and footer.  I'm not sure why its doing this since I'm not using float.  
image

styles/stylesheet.css
```diff
@@ -0,0 +1,627 @@
...
+.intro img {
+    max-width: 100%;
+    height: auto;
+}
+
+.dogs,
+.all-dogs {
+    display: grid;
+    grid-template-columns: repeat(auto-fit, minmax(0, 32%));
+    grid-gap: 1% 1%;
```
tylervanbrocklin: Think this is what is causing you some trouble. Instead of using a % width, this is a case where setting a constant px size works ok. Try changing to: `grid-gap: 20px;`. Should prevent your content from overlapping

js/blog.js
```diff
@@ -0,0 +1,61 @@
...
+   'blogHeader':'Teach Your Dog to Fetch',
+   'blogP1':'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Pellentesque blandit lectus non nisl feugia tristique. Cras massa purus, pharetra at diam vel, tincidunt volutpat metus. Cras sed rutrum dolor, id maximus felis. Nullam pulvinar ex non lectus tempus, id euismod lectus ullamcorper. Sed eleifend augue et libero faucibus semper. Curabitur dapibus enim elit, in molestie nisl euismod vel. In efficitur lacinia lorem, suscipit blandit est volutpat nec.',
+   'blogP2':'Nunc vitae libero laoreet, elementum lacus sed, ultricies nibh. Nam elementum commodo erat, lobortis iaculis justo elementum eget. Mauris id lacus massa. In maximus elit euligula cursus consequat. Nullam a massa eu mi fringilla ornare sodales non augue. Lorem ipsum dolor sit amet, consectetur adipiscing elit. In a dolor ut nulla mattis eleifend id a metus. In porta orci nec pellentesque ullamcorper. Phasellus sodales ipsum et lacus eleifend, nec sodales lacus aliquet.'
+  }
+];
+
+
+for (let i = 0; i<blogContent.length; i++) {
+    
+// each blog is assigned to an aritcle which are enclosed in a container
```
tylervanbrocklin, line 31: Nice work on the dynamically generated posts. This looks great! Parsing through data and generating HTML elements is used everywhere in web. You'll do some more of this in the next course as well with vue.

blog_copy.html
```diff
@@ -0,0 +1,148 @@
...
+            <h1>Seattle's premiere dog adoption agency</h1>
+        </section> <!-- end of hdr-left -->
+
+        <section class="hdr-right">
+            <div class="cart">
+                <img class="cart" src="images/cart-icon.svg" alt="image of a cart">
+                <p>$1234.56</p>
+            </div>
+            <!--end of cart  -->
+            <button type="submit" value="Submit" name="submit">Checkout</button>
```
tylervanbrocklin, line 25: Your checkout button should link you to the checkout page

js/form.js
```diff
@@ -0,0 +1,54 @@
...
+    //     } 
+    // }
+
+
+});
+
+// console log the form input using jQuery
+$('button').on('click', function (event) {
+    event.preventDefault();
+    let values = $('form').serialize();
```
tylervanbrocklin: We are wanting you to console.log each variable separately, or build a javascript object and print it out. Below is an example of what you'd be logging if it were an object. You are welcome to use jQuery as well, but we'd like to see you access each variable individually 
`{
     name: 'name',
     location: 'kent'
}`

js/dogs-javascript.js
```diff
@@ -0,0 +1,68 @@
...
+
+$('.clear-more-dogs').click(function(){
+  $('.total').text('$0.00');
+});
+
+}
+
+// change the border on dog tiles to Green and background to light blue when user mouses over the article
+$('article').on('mouseover', function (e) {
+  $(this).removeClass('noshow');
```
jimmynono, line 32: When you addClass it might override what you have, so it might not be necessary to remove the old class. Kind a nice thing about jQuery

PR #73, opened 2021-01-20
Assignment 1

dcp6:
Added all pages, 1-4. Styled and validated.


blog
```diff
@@ -0,0 +1,154 @@
...
+<body>
+
+<header>
+<div class="header-holder">
+<section class="left-side-header">
+<img src="images/adoptadog-logo.svg">
+<p>Seattle's premiere dog adoption agency</p>
+</section>
+<section class="right-side-header">
+  <p><i class="fas fa-shopping-cart"></i> $1234.56  </p>
```
jimmynono, line 20: You'll want to use the asses we've provided. You can use SVG's just like an image tag. You should take out the font awesome out of your code, we're doing thins the hard way. =)

blog
```diff
@@ -0,0 +1,154 @@
...
+  <button> Checkout</button>
+    
+    </section>
+</div>    
+</header>
+<nav>
+<ul class="header-nav-list-holder">
+    <li class="header-nav-list-item"><a href="#home">Home</a></li>
+    <li class="header-nav-list-item">/</li>
```
jimmynono, line 29: These you can handle with css _pseudo_ elements to inject things like this _after_ elements. See if you can't figure out how.

blog
```diff
@@ -0,0 +1,154 @@
...
+    <ul>
+        <li><a href="#home">Home</a></li>
+        <li>/</li>
+        <li><a href="#dogs">Dogs</a></li>
+        <li>/</li>
+        <li><a href="#contact">Contact</a></li>
+        
+        </ul>    
+     </nav>
+    <p>&#169; 2018 Adoptadog, LLC | Seattle, WA</p>
```
jimmynono, line 144: you can also use `copy;` to show the copy right symbol. It's better to use that syntax in your HTML

blog.html
```diff
@@ -0,0 +1,111 @@
...
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+<link rel="stylesheet" type="text/css" href="styles.css">  
+<link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.15.1/css/all.css" integrity="sha384-vp86vTRFVJgpjF9jiIGPEEqYqlDwgyBgEF109VFjmqGmIY/Y4HV4d3Gp2irVfcrp" crossorigin="anonymous">
+  <title>Blog</title>
+</head>
+
+<body>
+
+<header>
+<div class="header-holder">
```
jimmynono: make sure to watch your indenting.

blog.html
```diff
@@ -0,0 +1,111 @@
...
+    </ul>
+    
+    </nav>    
+<!--<div id="wrapper">-->
+<h2 class="page-header">Adoptadog Blog</h2>
+
+    <div class="grid-item-container"> 
+    
+<section class="blog-main-holder">
+<h3 class="hidden-header">Main Blog Holder 1</h3>
```
jimmynono: if you use a different HTML5 element for these blog _articles_ you won't need the hidden headers

checkout.html
```diff
@@ -0,0 +1,185 @@
...
+    <li class="header-nav-list-item">/</li>
+    </ul>
+    
+    </nav>    
+<!--<div id="wrapper">-->
+<h2 class="page-header">Checkout</h2>
+
+    <div class="grid-item-container"> 
+<form>
+    <label class="form-label" for="name">Name</label><br>
```
jimmynono: make sure all of your inputs and labels have matching id's and for's. There is at least one repeating and some missing.

blog
```diff
@@ -0,0 +1,154 @@
...
+    <li class="header-nav-list-item"><a href="#blogs">Blogs</a></li>
+    <li class="header-nav-list-item">/</li>
+    </ul>
+    
+    </nav>    
+<!--<div id="wrapper">-->
+<h2 class="dogs-list-header">Here are all our Dogs!</h2>
+    <div class="grid-item-container"> 
+
+<section class="dog-holder">
```
tylervanbrocklin, line 41: You'll want to be more consistent with your identation. It'll make your code much easier to use. Most code editors have an auto-ident feature that should align your code for you

blog
```diff
@@ -0,0 +1,154 @@
...
+    <img class="adoption-pic" src="images/murphy-card.jpg">
+     <h3>Murphy</h3>       
+    <p class="dog-cost">Cost to Adopt: $123.45</p> 
+    <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Cras maximus feugiat elementum. Pellentesque aliquet in lacus laoreet hendrerit. Proin eget tellus odio. Sed posuere sem eget blandit cursus. Curabitur venenatis, leo a malesuada sollicitudin, nunc diam luctus mauris, id semper urna massa maximus orci. </p>
+    <button class="adopt-button">Adopt</button>
+        </section>
+    <section class="dog-holder">
```
tylervanbrocklin, line 48: Sections should be used for distinct sections on your page. Like maybe a section for your intro text, then a section for dog tiles. In this case, you'll want to wrap each dog tile with a div instead of a section.

Also, I would suggest adding padding to your dog-holder so the content isn't squashed in the box

blog.html
```diff
@@ -0,0 +1,95 @@
...
+   <ul class="header-nav-list-holder">
+            <li class="header-nav-list-item "><a class="needs-slash" href="index.html">Home</a></li>
+            <li class="header-nav-list-item"><a href="dogs.html" class="needs-slash">Dogs</a></li>
+            <li class="header-nav-list-item"><a href="blog.html">Blogs</a></li>
+         </ul>
+      </nav>
+      <!--<div id="wrapper">-->
+      <h2 class="page-header">Adoptadog Blog</h2>
+      <div class="grid-item-container">
+         <article class="blog-main-holder">
```
tylervanbrocklin: I would suggest using flex for your blog-main-holder. So when pushed to mobile screen size, your text wraps underneath your image.

checkout.html
```diff
@@ -0,0 +1,168 @@
...
+      <h2 class="page-header">Checkout</h2>
+      <div class="grid-item-container">
+         <form>
+            <label class="form-label" for="name">Name</label><br>
+            <input class="form-box" type="text" id="name" name="name"><br>
+            <label class="form-label" for="email">Email</label><br>
+            <input class="form-box" type="email" id="email" name="email"><br>
+            <label class="form-label" for="address">Street Address</label><br>
+            <input class="form-box" type="text" id="address" name="address"><br>
+            <div class="city-state-zip-holder">
```
tylervanbrocklin, line 41: Use flexbox for this class. So on mobile, your city, state and zip can wrap to a new line.

styles.css
```diff
@@ -0,0 +1,250 @@
...
+a {
+    color:navy !important;
+    text-decoration:none;
+}
+a.needs-slash::after {
+    content: "/";
+    padding-left:10px;
+    margin-right:-5px;
+}
+a.needs-slash2::after {
```
jimmynono, line 84: You could probably find a way to combine these, since the padding/margins should be the same as they are in between elements

styles.css
```diff
@@ -0,0 +1,250 @@
...
+    background-color:white;
+    
+}
+*/
+.grid-item-container {
+    width:85%;
+    margin: 0px auto;
+    display:grid;
+    column-gap: 6.67%;
+    grid-template-columns: 20% 20% 20% 20%;
```
jimmynono: great! We'll learn how this could change at different widths.

styles.css
```diff
@@ -0,0 +1,250 @@
...
+    grid-template-rows: auto;
+ 
+}
+.initial-box {
```
jimmynono, line 119: this will behave better if you pull this out of your grid-item container.

styles.css
```diff
@@ -0,0 +1,250 @@
...
+h2.adopt-primary-header {
+    margin-bottom: 10px;
+}
+.intro-para {
+    line-height:35px;
+}
+.adoption-pic {
+    width:100%;
+}
+.initial-box img {
```
jimmynono, line 170: This image doesn't really scale well going narrower. There are a few ways you could do this. maybe image height fit-content?  Or.     width: 30%;  max-height: 100%; object-fit: contain; There are a few ways you could handle this

styles.css
```diff
@@ -0,0 +1,250 @@
...
+    display:flex;
+    flex-flow:row wrap;
+    justify-content:center;
+    align-items:center;
+}
+footer img {
+    width:100%;
+    display:block;
+}
+.blog-main-holder {
```
jimmynono, line 230: You have a grid property around your blog posts, that isn't really necessary.

styles.css
```diff
@@ -0,0 +1,250 @@
...
+}
+.location-holder {
+    margin-right: 20px;
+}
+.dog-checkout-holder-container {
+    display:flex;
+    flex-flow: row nowrap;
+}
+.dog-checkout-holder {
+    width: 450px;
```
jimmynono: Try to not use px for this kind of thing. Either use a percentage, or even a REM to do this. For this I'd recommend a percentage.

styles.css
```diff
@@ -0,0 +1,250 @@
...
+.checkout-cost {
+    margin-bottom: 20px;
+}
+.see-more-dogs-button-holder {
+    display: flex;
+    justify-content: center;
+    margin-top: 30px;
+}
+button.see-more-dogs-button {
+    height:40px;
```
jimmynono: translating these values into rem's would be good, so they scale with the font size

styles.css
dcp6: I used to grid properties to adjust the positions of things. Should I figure out an alternative?

jimmynono: It's just overkill. Use grid for big grid things, and most small components can be handled with some of the more basic things. I'd argue you could lay out these articles using just flexbox. Maybe even just simple display properties. 

Also `.grid-item-container` isn't really doing much with the grid. I turned off all the properties in the inspector and nothing changed.

myscripts.js
```diff
@@ -0,0 +1,64 @@
...
+
+}
+
+function dogInfoSheetReturn(dogCost,dogName, dogBreed) {
+    alert('Your dog cost ' + '$' + dogCost + ' and is named ' + dogName + ' of the breed ' + dogBreed + '.');
+}
+function blogFieldsGenerator() {
+        const gridItemContainer = document.getElementById('gridItemContainer');
+
+        for (i = 0; i < dogBlogData.length; i ++){
```
jimmynono, line 48: This is hard to read. I'd recomment a space before each const statement (other than the first) it breaks up the huge wall of text)

PR #79, opened 2021-01-21
Lesson 10 Assignment

KyruCreates:
Created HTML pages and set up some file structure and CSS files.


checkout.html
```diff
@@ -0,0 +1,117 @@
...
+            <li>/</li>
+            <li><a href="./dogs.html">Dogs</a></li>
+            <li>/</li>
+            <li><a href="./blog.html">Blog</a></li>
+          </ul>
+        </nav>
+      </header>
+
+      <article class="checkout" id="main-content">
+        <h2 id="main-content">Checkout</h2>
```
tylervanbrocklin: Note that IDs should be unique and only used in one HTML tag. If you'd like to use main-content in multiple places, you would want to convert this to a class. This will pop up in the code validator as well.

blog.html
```diff
@@ -0,0 +1,84 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+  <head>
+    <meta charset="utf-8">
```
jimmynono, line 4: You MUS have the meta tag mentioned in week one. Very important.

styles/stylesheet.css
```diff
@@ -0,0 +1,253 @@
+* {
+  box-sizing: border-box;
```
jimmynono, line 2: YES!

styles/stylesheet.css
```diff
@@ -0,0 +1,253 @@
...
+.checkout {
+  margin-top: 2rem;
+}
+.checkout div {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: baseline;
+}
+#cost {
```
jimmynono: For CSS use classes. It's ok if you only use a class once.

styles/stylesheet.css
```diff
@@ -0,0 +1,253 @@
...
+}
+button, .checkout-page input[type=submit] {
+  background-color: lightgreen;
+  color: white;
+  font-size: 1rem;
+  width: 8rem;
+  margin-right: 3rem;
+  border: none;
+  border-radius: 2px;
+  cursor: pointer;
```
jimmynono, line 231: YES!

styles/stylesheet.css
```diff
@@ -0,0 +1,253 @@
...
+#page-content section button {
+  width: 11rem;
+  height: 2rem;
+}
+#dogs {
+  display: block;
+  margin-left: auto;
+  margin-right: auto;
+  margin-top: 3rem;
+  margin-bottom: 3rem;
```
jimmynono, line 302: you can shorthand this to 

`margin: 3rem auto;`

styles/stylesheet.css
```diff
@@ -0,0 +1,253 @@
...
+  font-weight: bold;
+  margin-bottom: 1rem;
+}
+footer {
+  background-color: darkblue;
+  color: white;
+  width: 100%;
+  height: 20rem;
+  /* positioning */
+  position: relative;
```
jimmynono: You probably don't need this positioning

styles/stylesheet.css
```diff
@@ -0,0 +1,253 @@
...
+  list-style-type: none; /* removes bullet points */
+  margin: 0; /* removes browser defaults */
+  padding: 0 100px;
+  overflow: hidden;
+  background-color: lightgrey;
+  font-weight: bold; /* makes text bold */
+  width: auto;
+}
+.menu li {
+  float: left; /* alignment of nav li */
```
jimmynono: Don't use a float here. You could use `display-inline-block` instead

styles/stylesheet.css
```diff
@@ -0,0 +1,253 @@
...
+  text-align: center;
+  padding: 1rem 1rem 1rem 0;
+}
+footer li a {
+  color: white;
+}
+footer li a:hover {
+  color: lightgreen;
+}
+.clearfix::after {
```
jimmynono: If you get rid of the float you wont need this.

KyruCreates:
Adjusted some CSS styling

index.html
```diff
@@ -0,0 +1,132 @@
...
+          <li><a class="active" href="./index.html">Home</a></li>
+          <li>/</li>
+          <li><a href="./dogs.html">Dogs</a></li>
+          <li>/</li>
+          <li><a href="./blog.html">Blog</a></li>
+        </ul>
+      </nav>
+
+      <article class="main-content">
+        <div class="summary">
```
tylervanbrocklin: Would suggest using flexbox here, specifically flex-wrap. This will allow your text to wrap under the image on smaller screen sizes. Your html here is great!

styles/stylesheet.css
```diff
@@ -0,0 +1,254 @@
...
+}
+.sec-1 img, .sec-2 img, .sec-3 img {
+  height: 15rem;
+}
+/********** DOG LIST **********/
+.page-content {
+  grid-area: 2 / 1 / 3 / 4;
+  grid-gap: 3rem;
+  display: flex;
+  flex-flow: row wrap;
```
tylervanbrocklin: Your grid looks awesome. Well done getting the columns to scale on different screen sizes

blog.html
```diff
@@ -0,0 +1,85 @@
...
+          <li>/</li>
+          <li><a href="./dogs.html">Dogs</a></li>
+          <li>/</li>
+          <li><a href="./blog.html">Blog</a></li>
+        </ul>
+      </nav>
+
+      <h2 class="blog" id="main">Adoptadog Blog</h2>
+      <article class="main-content">
+        <div class="sec-1">
```
tylervanbrocklin: Your HTML here is perfect. I would suggest using flexbox here to allow your text to wrap under the image. Also, you shouldn't need to create a separate class for each article (sec-1, sec-2) one class should be able to be applied for each article.

checkout.html
```diff
@@ -0,0 +1,125 @@
...
+          <li><a class="active" href="./index.html">Home</a></li>
+          <li>/</li>
+          <li><a href="./dogs.html">Dogs</a></li>
+          <li>/</li>
+          <li><a href="./blog.html">Blog</a></li>
+        </ul>
+      </nav>
+
+      <article class="checkout-page">
+        <h2 class="check" id="main">Checkout</h2>
```
tylervanbrocklin, line 40: Your form looks good. Maybe see if you can have your text fields scale. So on desktop they take up more space, then on mobile they become compact like what you have now.

index.html
```diff
@@ -0,0 +1,134 @@
...
+            <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam gravida nec ligula lobortis tincidunt. In faucibus nisi quis orci imperdiet, a suscipit odio tristique.</p>
+            <button type="button" name="adopt" onclick="alert('Your total is: ' + totalCost(123.45))">Adopt</button>
+          </section>
+
+          <section>
+            <img src="./images/jack-card.jpg" alt="A picture of Jack, a brown and white Beagle" onclick="alert('This is ' + dogInfo('Jack', 'a Beagle', 123.45))">
+            <h3>Jack</h3>
+            <p><strong>Cost to Adopt:</strong> $123.45</p>
+            <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam gravida nec ligula lobortis tincidunt. In faucibus nisi quis orci imperdiet, a suscipit odio tristique.</p>
+            <button type="button" name="adopt" onclick="alert('Your total is: ' + totalCost(123.45))">Adopt</button>
```
tylervanbrocklin: Generally not a great practice to have two function calls in the onclick. You typically want to keep as little JS as possible in your HTML code. What would be better is to just have `totalCost(123.45)` and then make the alert in your function

styles/stylesheet.css
```diff
@@ -0,0 +1,296 @@
...
+body a {
+  text-decoration: none; /* removes hyperlink underlines */
+}
+body a:hover {
+  color: black;
+  text-decoration: underline;
+}
+/********** MEDIA QUERIES **********/
+/* for phones - dog grid layout needs to change from 3 columns to 1 (home) and hide blog aside */
+@media not print and (min-width: 576px) and (max-width: 767px) {
```
tylervanbrocklin: I suggest changing all of your media queries from `@media not print` to `@media screen`. You should then notice your blog preview going away on mobile. I would also suggest removing the minimum width here. There may be mobile devices smaller than 576px. For example, if I stretch my browser window to the smallest, it'll go below your min width here.

styles/stylesheet.css
```diff
@@ -0,0 +1,296 @@
...
+    margin: -1rem 6rem 2rem 4rem;
+  }
+  .page-content {
+    grid-gap: 2rem;
+    display: flex;
+    flex-flow: row wrap;
+    margin-left: 2rem;
+  }
+  .page-content-dogs {
+    grid-area: 2 / 1 / 3 / 4;
```
tylervanbrocklin: Your dog tiles on your home page don't appear to be scaling. I suggest looking at using css grid, specifically auto-fit and minmax for the tile sizes. See MDN doc here: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Auto-placement_in_CSS_Grid_Layout

styles/stylesheet.css
```diff
@@ -0,0 +1,370 @@
...
+    width: 13rem;
+    padding-right: 1px;
+  }
+  .page-content-dogs section button, .page-content section button {
+    width: 11rem;
+    height: 2rem;
+  }
+}
+/* for tablet - dog grid layout needs to change from 3 columns to 2 (home) */
+@media screen and (min-width: 768px) and (max-width: 991.9999px) {
```
jimmynono: This doesn't quite work. I can't see your blog previews at 768. I think your max-width is causing wierdness. Generally you won't need to combine max and min width. You could do a media quiery at 768, then another at 992px just using min-width and as long as the larger value is below the smaller value in css it'll sort itself out.

scripts/blogpost.js
```diff
@@ -0,0 +1,163 @@
...
+  'Uptiusd andesci qui nem aut vendion ectur? Debis que explaut laborenia que doluptur, con et labor abor sant poreperum dio quat que doluptatur aut voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam voluptassim quisciatquam ea ad qui con nes cus esere dolut hicto teni solutenis alit ulparume nonseca estorer spernam.'
+];
+
+const posts = [
+  'post1',
+  'post2',
+  'post3'
+];
+
+for (let i = 0; i < posts.length; i++) {
```
tylervanbrocklin: You are very close on this one. Instead of making three arrays (post1, post2, post3), you will want your posts to be javascript objects. Here's an example of one:
```
{
     title: "The Title",
     image: "./images/blog-1.jpg",
     para-one: "The first paragraph text here"
}
```

Now if we turn that into an array:

```
[{
     title: "The Title",
     image: "./images/blog-1.jpg",
     para-one: "The first paragraph text here"
},
{
     title: "The Title",
     image: "./images/blog-1.jpg",
     para-one: "The first paragraph text here"
},
{
     title: "The Title",
     image: "./images/blog-1.jpg",
     para-one: "The first paragraph text here"
}]
```

Then we can loop through the array and access elements like this:

```
for(let i = 0; posts.length; i++) {
let blogTitle = document.createElement("h3");
blogTitle.innerHTML = posts[i].title
}
```

PR #96, opened 2021-02-10
redo github

jgalber:


scripts/script.js
```diff
@@ -0,0 +1,53 @@
+ 
+function cost() {
+    alert("$123.45");
+}
+
```
tylervanbrocklin, line 5: You are on the right track here, but not quite what we're looking for. Remember, we want to show the cart total. So if you click two dogs, it should display $246.90". You can do this by creating a variable to keep track of cost.

tylervanbrocklin: Also, you'll want to use function parameters. So something like: `func total(cost, dog, breed)` then when you call the function in your HTML code, you'd pass in the values.

scripts/script.js
```diff
@@ -0,0 +1,53 @@
+ 
+function cost() {
+    alert("$123.45");
+}
+
+function infoMurphy(){
+    alert("Murphy \nBreed: Boxer \n\nAdoption fees: $123.45");
+}
```
tylervanbrocklin, line 11: Functions are designed to be reused. In this case, you should only have one info function that works for all dogs. You'd do this by creating a function with info: `func total(cost, dog, breed)`

scripts/script.js
jgalber: Thank you so much Tyler, I have already pulled the project again with the committed changes.

scripts/script.js
jgalber: Great!! now, that makes sense, less and more effective code. As you suggested I change the name of the function and add some values, parameters and arguments, to wrap the function. Forgot to mention the sum of the totals when clicking the 'adopt' button. Thanks for your help.

styles/style.css
```diff
@@ -0,0 +1,1147 @@
...
+
+small {
+    color: #fff;
+    margin-top: 1rem;
+    padding-bottom: 4rem;
+}
+
+
+/* MEDIA QUERIES */
+@media screen and (min-width: 0px) and (max-width: 576px) {
```
tylervanbrocklin: Great work with the media queries. Note you do not need to specify a min-width here. If you just set a max-width, the compiler will allow for anything under your max-width.

styles/style.css
```diff
@@ -0,0 +1,1147 @@
...
+    .blog__heading-one {
+        font-size: 3rem;
+        text-align: center;
+    }
+}
+
+
+@media screen and (max-width: 576px) {
+    .blog_content {
+        display: none;
```
tylervanbrocklin: nicely done getting the blog content to disappear on small screen sizes

styles/style.css
jgalber: Thanks Tyler. I was honestly not sure enough to define the 'minimum width' with the value of 0. I thought this idea will keep the layout strictly accurate to the specified style with the root value of 0. So some best practices recommend 'minimum widths' rather than 'maximum widths', but it makes sense here. Thanks again for your help, sometimes simpler is better.

scripts/main.js
```diff
@@ -0,0 +1,81 @@
...
+        content: 'Puppy kitty ipsum dolor sit good dog foot stick canary. Teeth Mittens grooming vaccine walk swimming nest good boy furry tongue heel furry treats fish. Cage run fast kitten dinnertime ball run foot park fleas throw house train licks stick dinnertime window. Yawn litter fish yawn toy pet gate throw Buddy kitty wag tail ball groom crate ferret heel wet nose Rover toys pet supplies.',
+        
+        content2: 'Pet Food pet supplies gimme five puppy cage food feathers food heel feathers running pet gate walk lazy dog Spike. Good Boy park lazy dog walk kibble Scooby snacks licks canary. Maine Coon Cat walk catch water dog slobber chew scratcher ID tag litter tuxedo dog house lazy cat park.',
+        
+        alt: 'a siluette of a teenage girl holding a ball and a dog ready to play in the sunset'
+    }
+    
+];
+
+    for (let i = 0; i < fields.length; i++) {
```
tylervanbrocklin, line 52: Great work dynamically generating the blog posts. In the next course in the series, vue.js will make this much easier. Your code looks great

scripts/form.js
```diff
@@ -0,0 +1,41 @@
+const form = document.getElementById('form');
+
+form.addEventListener('submit', function(event){
```
tylervanbrocklin, line 3: Your form outputs in console and displays an alert as required. Nice job!

PR #99, opened 2021-02-11
Adopt A Dog Project (Yuting Lin)

filizlin:


filizlin:
Assignment 04 javascript for Adopt button

blog.html
```diff
@@ -0,0 +1,112 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+
+<head>
+  <meta charset="utf-8">
+  <title>>Adopt A Dog</title>
+
```
jimmynono, line 7: You will need that Meta tag we referenced in week on on ALL HTML files

blog.html
```diff
@@ -0,0 +1,112 @@
...
+<body>
+  <header>
+    <div class="row">
+      <div class="col span_4_of_12">
+        <img class="header-logo" src="./images/adoptadog-logo.svg" alt="adopt a dog logo">
+        <h1>Seattle's Premiere dog adoption agency</h1>
+
+      </div>
+      <div class="col span_7_of_12">
```
jimmynono, line 20: If you're going to write your own grid, see if you can write an offset that you could apply to the next column instead of an extra grid.

That being said you could layout  this whole header with flexbox

blog.html
```diff
@@ -0,0 +1,112 @@
...
+
+      </div>
+    </div>
+  </header>
+
+  <main>
+    <div class="row span_12_of_12 headernavigation">
+      <nav>
+        <a class="headernavigationitem" href="./index.html">Home</a>
+        <span> / </span>
```
jimmynono, line 39: See if you can't figure out how to do this with a pseudo element in CSS

styles/blog.css
```diff
@@ -0,0 +1,188 @@
...
+
+.headernavigation {
+    float: left;
+    display: block;
+    text-align: left;
+    align-content: center;
+    list-style-type: none;
+    overflow: hidden;
+    color: #083B60;
+    padding-left: 15%;
```
jimmynono, line 41: for most padding and margin I like rem's (or px) things can look really odd at small widths

styles/blog.css
```diff
@@ -0,0 +1,188 @@
...
+
+.check-out {
+  background-color: #81B741;
+  color: white;
+  padding: 10px 4px;
+  text-align: center;
+}
+
+.headernavigation {
+    float: left;
```
jimmynono, line 34: There's a better way than float to lay out these items. Float should be rarely used

styles/blog.css
```diff
@@ -0,0 +1,188 @@
...
+
+.footernavigation a {
+    color: white;
+    text-decoration: none;
+    padding: 5px;
+  }
+
+/* grid layout setting */
+.col {
+   float: left;
```
jimmynono, line 111: Yeah. Don't use float for grids. Find a more modern approach using flexbox or css grid

filizlin: I tried to avoid any grid layout but had to put it back to make the responsive design work, and with the grid I couldn't figure out how to avoid floating them for the layout.  I will be sure to think about this more and see if I can find a solution to this!

filizlin:
Assignment06

filizlin:
Assignment07

styles/index.css
```diff
@@ -0,0 +1,428 @@
...
+}
+
+/* skeleton */
+header {
+  clear: both;
+  margin-bottom: 0;
+}
+
+main {
+  margin-left: 15rem;
```
jimmynono: this is too large and makes your site pretty squished. I'd take it out entirely.

index.html
```diff
@@ -0,0 +1,169 @@
...
+  <title>Adopt A Dog</title>
+
+  <link rel="stylesheet" type="text/css" href="./styles/index.css">
+
+
+</head>
+
+<body>
+  <header class="col-12">
+    <div class="header-banner">
```
tylervanbrocklin, line 15: Would recommend using flexbox wrap, so on mobile screens the checkout button wraps below the logo

styles/index.css
```diff
@@ -0,0 +1,428 @@
...
+}
+
+footer {
+  clear: both;
+  margin-bottom: 0;
+}
+
+/* header setting */
+.header-banner {
+  background-image: linear-gradient(45deg, blue, grey);
```
tylervanbrocklin, line 32: nice work with the gradient! now if the image doesn't load, it'll just render a gradient!

filizlin:
Assignment08

filizlin:
Milestone 3 Assignment

filizlin:
Assignment 10

javascript/adoptbutton.js
```diff
@@ -0,0 +1,14 @@
+let Final = 0;
+
+function AddtoCart(value) {
```
tylervanbrocklin, line 3: very small, but usually for function and variable names we use camel case. so start out with lowercase and capitalize each word after, so would be: `function addToCart(value)`

javascript/adoptbutton.js
```diff
@@ -0,0 +1,14 @@
...
+    Final = Final + value;
+    console.log("Adding " + value + "to cart.  The total adoption fee is now: " + Final);
+    $("#total").text("$ " + Final);
+};
+
+function Dog(Name,Breed,Fee) {
```
tylervanbrocklin, line 9: Parameter names should also utilize camel case, so in this case, all lowercase

lib/jquery.js
```diff
@@ -0,0 +1,25 @@
+//A hover effect that gives the tile an outline, or give it a box shadow.
+$(".dogs").hover(function(){
+    $(this).addClass("hoverborder");
+}, function() {
+    $(this).removeClass( "hoverborder");
+  }
+);
+
+//Zooms in on the dog photo
+$(".dog-image").hover(function(){
```
tylervanbrocklin, line 10: Nice job with the effects. Would see if you could mess around with z-index in your hoverborder class, so the image is always on top. Sometimes it ends up behind the other dog tiles

PR #107, opened 2021-03-04
Lesson-07

dseaberg:


blog-page.html
```diff
@@ -0,0 +1,94 @@
...
+          <li><a href="adopt-page.html">Dogs</a></li>
+          <li><a href="blog-page.html">Blog</a></li>
+        </ul>
+      </nav>
+    </header>
+
+    <main>
+      <section>
+        <h2>Adoptadog Blog</h2>
+        <article class="blog">
```
tylervanbrocklin: one suggestion would be to use flexbox wrap for your blog class. this will allow your text to wrap below the image on smaller screen sizes. Right now your text gets squished on small screens

PR #110, opened 2021-03-11
Lesson-08

dseaberg:


scripts/script.js
```diff
@@ -0,0 +1,137 @@
...
+  console.log('Adding blog elements');
+  const section = document.getElementById('blog-page-content');
+  if (section !== null) {
+    for (let i = 0; i < blogsToAdd.length; i++) {
+      section.appendChild(blogsToAdd[i]);
+    }
+  }
+}
+
+function createBlog(blogData) {
```
jimmynono, line 155: nice to separate this into it's own function!

scripts/script.js
```diff
@@ -0,0 +1,137 @@
...
+  const blogsToAdd = [];
+  for (let i = 0; i < allBlogData.blogs.length; i++) {
+    const blog = createBlog(allBlogData.blogs[i]);
+    blogsToAdd.push(blog);
+  }
+
+  console.log('Adding blog elements');
+  const section = document.getElementById('blog-page-content');
+  if (section !== null) {
+    for (let i = 0; i < blogsToAdd.length; i++) {
```
jimmynono, line 149: You do loop through two things. You could probably simplify this add Blogs function to only call one loop (the blog data)

PR #112, opened 2021-03-18
Lesson-09

dseaberg:


scripts/script.js
```diff
@@ -0,0 +1,227 @@
...
+function showDogInfo(dogName, dogBreed, adoptionFee) {
+  let infoMessage = 'Name: ' + dogName + '\nBreed: ' + dogBreed + '\nAdoption Fee: ' + adoptionFee;
+  alert(infoMessage);
+}
+
+/* Add the dog's adoption fee to the total price. */
+let totalPrice = 0;
+function addDogToTotal(adoptionFee) {
+  totalPrice += adoptionFee;
+  $('#cart-total-main-page').text('$' + totalPrice);
```
tylervanbrocklin, line 11: Would suggest using rounding. I clicked a few tiles and the total price has multiple decimals. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round

PR #190, opened 2021-12-05
lesson-01-assignment branch; submission for lesson 4, assignment 2

bit-encoder:
Resubmitting pull request.


bit-encoder:
Attempted best to copy from .pdf file of exercise. Any chance to provide what fonts  and font sizes are used in the .pdf?

This has also been optimized for small screens, 1080x1920 pixels.

